PR #34, opened 2014-01-30
tests: Fix undefined reference to `ck_assert_ptr_eq`

### LINKED ISSUE #33: tests: undefined reference to `ck_assert_ptr_eq'
The ck_assert_ptr functions were added in version 0.9.10 of check (see http://sourceforge.net/p/check/code/738/tree//trunk/NEWS). The version of check available in Debian stable is 0.9.8:

```
$ cat /etc/debian_version
7.3
$ apt-cache show check |head -n2
Package: check
Version: 0.9.8-2
```

Would you accept a patch changing `ck_assert_ptr_eq(NULL, str)` to `ck_assert(NULL == str)`?

atj:
This PR resolves #33.

The ck_assert_ptr family of functions were added in check 0.9.10, which
unfortunately is newer than the versions available in the current stable
releases of Debian and Ubuntu. We workaround this by replacing calls to
`ck_assert_ptr` with `ck_assert`. The failure messages won't be as
pretty but at least the tests run...

Signed-off-by: Adam James adam.james@transitiv.co.uk

### DIFF
--- a/tests/test-checks.c
+++ b/tests/test-checks.c
@@ -28,8 +28,8 @@ START_TEST(one_line_no_perfdata)
 	output = strdup(full_output);
 	parse_check_output(output, &short_output, &long_output, &perf_data, FALSE, FALSE);
 	ck_assert_str_eq(short_output, full_output);
-	ck_assert_ptr_eq(NULL, long_output);
-	ck_assert_ptr_eq(NULL, perf_data);
+	ck_assert(NULL == long_output);
+	ck_assert(NULL == perf_data);
 
 }
 END_TEST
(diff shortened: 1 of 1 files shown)


vincentbernat:
You could define `ck_assert_ptr_eq` as this:

```c
#if (CHECK_MAJOR_VERSION == 0 && (CHECK_MINOR_VERSION < 9 || (CHECK_MINOR_VERSION == 9 && CHECK_MICRO_VERSION < 10)))
# define ck_assert_ptr_eq(X,Y) do {					\
		void* _ck_x = (X);					\
		void* _ck_y = (Y);					\
		ck_assert_msg(_ck_x == _ck_y,				\
			      "Assertion '"#X"=="#Y"' failed: "#X"==%p, "#Y"==%p", \
			      _ck_x, _ck_y);				\
	} while (0)
#endif
```

PR #66, opened 2014-05-20
Fix shell command parsing

philrandal:
Valid shell commands of the form

command_line VAR='VALUE' /bin/something

were not properly parsed by runcmd.c.

This caused issues with check_mk's flexible notifications.

This patch fixes the issue by detecting an '=' character before the first space in a command and forcing the use of a command shell to process it, rather than using execv

See http://tracker.nagios.org/view.php?id=582 for the full details

### DIFF
--- a/naemon/commands.c
+++ b/naemon/commands.c
@@ -1516,6 +1516,7 @@ static int del_downtime_by_filter_handler(const struct external_command *ext_com
 	}
 }
 
+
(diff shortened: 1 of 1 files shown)


tvestelind:
This PR has been accepted (with slight modification, some `break` were missing the switch-case).

Thank you @philrandal !

PR #92, opened 2014-10-29
Service heredity of contacts and contact groups

nook24:
The problem is, that due to the following if condition, a service only
can inherit contact groups AND contacts:
if(temp_service->have_contact_groups == FALSE &&
temp_service->have_contacts == FALSE) {

But if you define a host with a contact and a contact group, and a
service with only a contact for example, only your service contact will
get an notification, but not the contact group that normally should be
inherited from the host.
If you want to stop the heredity you can add a contact or contact group
with notification interval none. I guess this is a better solution with
more flexibility for the admins

### DIFF
--- a/naemon/xodtemplate.c
+++ b/naemon/xodtemplate.c
@@ -3482,11 +3482,18 @@ static int xodtemplate_inherit_object_properties(void)
 			continue;
 
 		/*
+		 * if the service has no contact groups specified, it will inherit
+		 * them from the host
+		 */
+		if(temp_service->have_contact_groups == FALSE) {
+			xod_inherit_str(temp_service, temp_host, contact_groups);
+		}
+		
+		/*
 		 * if the service has no contacts specified, it will inherit
 		 * them from the host
 		 */
-		if (temp_service->have_contact_groups == FALSE && temp_service->have_contacts == FALSE) {
-			xod_inherit_str(temp_service, temp_host, contact_groups);
+		if(temp_service->have_contacts == FALSE && temp_host->have_contacts == TRUE) {
 			xod_inherit_str(temp_service, temp_host, contacts);
 		}
 


nook24:
Hi guys, any update for this?

ageric:
Sorry. Birthday parties and christmas celebrations got in the way.

The idea behind only inheriting objects that can be chosen as groups or as single objects is that is that the actual target object (in this case contacts) are handled as a unit no matter how they're selected.

This was a deliberate change made several years ago in response to user requests, so I'm reluctant to change it now. As you say, one can work around unwanted inheritance by specifying contacts with notification_period none, but that is hacky and ugly. With the current code, you can get what you want by specifying the contactgroups for the service as well, which is logical and makes sense.

I suppose we could add some sort of special marker name that means "inherit this value from hosts no matter what we have in contacts" or something, but I'm really against magic constants in config files. They tend to require a lot of documentation, not get used much and break often in unexpected ways.

nook24:
The contact group with notification_period none is a little bit hacky thats true, but is guess this is how inheritance works on a file system for example or on CSS. So everything gets inherited as long as nothing overwrite this value.
So on a file system users and user groups are not chained together. And i guess this is the same for contact and contact groups.

As far as i know, this was changed with one of the Nagios 4 versions, but i'm not sure.

For me it was very easy to write the config files for this inheritance logic.

tvestelind:
Hi,

Sorry, this PR is quite old but better late than never...

From our (OP5's) part we'd rather not have the change either. The reason for this is the same as @ageric has stated, if you add a contact or a contact group to a service that is to be regarded as an exception and only what you explicitly add shall have an effect.

nook24:
Hi @tvestelind thanks for your reply. I guess we are all used to the new behavior now, so we can close this PR.

PR #94, opened 2014-11-14
Cleanup event queue, and checks

pengi:
There are lot of stuff in the check lifecycle which just asks for trouble... The event queue is not generic.

Also the event queue internally is converted to monotonic time.

This pull request needs review, and probably a couple of fixes too. But it would be great to start reviewing this stuff.


catharsis:
These changes are already merged. Closing this PR.

PR #102, opened 2015-03-07
Updated daemon-init.in to protect reloads from killing the daemon

starvingprogrammer:
On CentOS 6, doing service reloads when having broken object configuration will kill the main naemon process.  Added check_config to the reload command will prevent the reload from happening and show that the current configuration is invalid.

### DIFF
--- a/daemon-init.in
+++ b/daemon-init.in
@@ -168,6 +168,8 @@ restart() {
 }
 
 reload() {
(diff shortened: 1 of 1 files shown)


sni:
still a good idea, but the main problem is that the core dies. i will create a seperate issue for that.

PR #109, opened 2015-04-15
Fix segfault issue on newline output from plugin

simenaasland:
Prevent Naemon from segfaulting when a plugin's
output is one or more newlines only.


oranki:
I can confirm this patch works. Applied the patch against git HEAD and tested with a dummy plugin:

```bash
#!/bin/bash
echo
```

Without the patch, Naemon segfaults.

PR #111, opened 2015-05-08
shadownaemon: use correct request index

simenaasland:
Since the removal of external commands, shadownaemon was
using the incorrect request index when updating last_request_count.
Consequently, the poll interval was decided based on wrong values.


sni:
thanks

PR #114, opened 2015-05-18
shadownaemon: faster return to short poll interval

simenaasland:
If there have been requests while using the long poll
interval, shadownaemon no longer sleeps for the remainder
of the long interval before returning to the main refresh loop.


sni:
thanks a lot

PR #122, opened 2016-01-31
Fix custom variables persistence & long_plugin_output behavior change

julthomas:
Hi,

The first patch fixes custom variables persistence across restarts. It is applicable to Nagios as well, at least from 3.2.3.

The second fixes a change of behavior regarding newlines in long_plugin_output. I started by trying the parse_check_output() from Nagios currnt master but the test suite did not pass anymore; I think there were 2 segfaults. So I came up with that using previous code from Nagios to handle <escape_newlines_please> and <newlines_are_escaped>.

Cheers,
Julien

### DIFF
--- a/src/naemon/checks.c
+++ b/src/naemon/checks.c
@@ -2923,12 +2923,41 @@ struct check_output *parse_output(const char *buf, struct check_output *check_ou
 /* parse raw plugin output and return: short and long output, perf data */
 int parse_check_output(char *buf, char **short_output, char **long_output, char **perf_data, int escape_newlines_please, int newlines_are_escaped)
 {
+	int x = 0;
+	int y = 0;
 	struct check_output *check_output = nm_malloc(sizeof(struct check_output));
+
+	/* We should never need to worry about unescaping here again. We assume a
+	 * common internal plugin output format that is newline delimited. */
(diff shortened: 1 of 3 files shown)


julthomas:
The travis build fails but looks like it's an issue with the build system :

`E: Unable to locate package gperf`
`The command "sudo apt-get install gperf" failed and exited with 100 during .`

catharsis:
I disagree with reintroducing those parameters in your second commit (obviously, since I removed them in the first place). The op5 fix for what I believe is the same issue you're experiencing is here: http://git.op5.org/gitweb?p=monitor/naemon.git;a=commit;h=2a0508615d5a48ff7bfbf26dcc3a883140acd18a

julthomas:
Hi Anton,

Thanks very much for your input on that. I did not thought about checking out the op5 repository. So this was done on purpose, okay.

Correct me if I'm wrong, I believe it changes livestatus (using neamon-livestatus 1.0.3) output/behavior when querying the long_plugin_output column because you can end up with several lines instead of a single one. What do you think about it ? Would you recommend to patch the livestatus side in order to add escaping back there?

Cheers,
Julien

julthomas:
I will resubmit a new pull request with only the custom variables persistence patch.

PR #123, opened 2016-02-13
xrddefault: Fix custom variables persistence accros restarts

julthomas:
This fix is not Naemon specific, it is also applicable to Nagios.


pengi:
The patch does, to some extent, make sense. But the code itself isn't really clear.

Can you try to clarify the actual bug you're trying to resolve? Preferably in the commit message?

sni:
The thing here is, that naemon writes changed custom variables prefixed with an option value into the retention.dat like
```
_VARNAME=1;value
```
On startup it then checks if the lenght of `1;value` is more than 3 which in fact skips changed custom variables if they only contain a single character or nothing at all.
So +1 for this PR.

julthomas:
Thank you Sven, that's exactly it.
I've amended the commit message.

PR #126, opened 2016-02-29
Don't make assumptions about size_t

catharsis:
In particular, don't assume it is an unsigned long, because on i386 (for
instance), it's not.

Signed-off-by: Anton Lofgren <alofgren@op5.com>


sni:
thanks

catharsis:
test comment, please ignore :)

PR #142, opened 2016-06-10
remove temp_file usage from retention.dat and status.dat file creation

stromsoe:
Make xrddefault_save_state_information() use retention_file as pattern to mkstemp()
Make xrddefault_save_state_information() check ferror() at file close and abort my_rename() if there are any errors
Make xsddefault_save_status_data() use status_file as pattern to mkstemp()
Make xsddefault_save_status_data() check ferror() at file close and abort my_rename() if there are any errors
Modify my_rename() to check for errno==XDEV and remove calls to my_fcopy()
Remove now unused my_fcopy() and my_fdcopy()


catharsis:
Thanks a lot!

LGTM, I've submitted this patch for internal review.

PR #146, opened 2016-07-29
Checking for Git binary

ipstatic:
In some build systems, git is not present unless defined as a build
dependency. However the .git directory may still be copied in along with
the source. This change will still allow the required behavior in case
.git is present but the Git binary is not.


sni:
Looks good to me.

pengi:
Merged

PR #147, opened 2016-08-26
Only reload with systemd if configtest is ok

eythori-sensa:
Configuration errors followed by a reload causes the server to stop without an error in systemd.
Adding a check before running the reload.


eythori-sensa:
Further checks show this does not work :( Closing this pull request

PR #166, opened 2017-01-19
remove NICELEVEL from systemd service

sni:
NICELEVEL is not set by default, so systemd constantly complains about the value not beeing set.


sni:
well, its commented in `/etc/sysconfig/naemon` as well, so i didn't want to completly remove it.

tvestelind:
Ok, that makes sense.

tvestelind:
One other reason I think it should be removed is that NICE is documented in systemd.unit while the sample config configuration values are not documented anywhere else, thus it's fine if those are commented.

PR #170, opened 2017-03-24
Minimize the amount of verify-config

mattiasr:
In a large setup where it can take more than 30 seconds to verify a config.

I propose a implementation similar to other systems, like Apache httpd.
* If you change your config and restart Apache httpd, the server won't start
and you have a broken config. (Expected)
* If you run a verify on the config, it will verify without interrupt with
your current running instance.

But doing this we can speed up the actual start/stop/restart of the naemon
process.

Signed-off-by: Mattias Ryrlén <mattiasr@op5.com>

### DIFF
--- a/daemon-init.in
+++ b/daemon-init.in
@@ -62,15 +62,6 @@ status() {
 start() {
     test -x $exec || exit 5
     test -f $config || exit 6
-    if [ "$checkconfig" = "false" ]; then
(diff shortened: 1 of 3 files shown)


sni:
Looks good to me

fmikker:
Nice. We should remember to update the documentation accordingly if needed.

PR #172, opened 2017-04-04
fix segfault in notifications when there is no eventbroker

sni:
nebcallback resultset may be null if eventbroker options prevent
notifications from beeing brokered. Do not segfault in those
cases.

### DIFF
--- a/src/naemon/nebmods.c
+++ b/src/naemon/nebmods.c
@@ -369,6 +369,8 @@ static neb_cb_resultset *neb_cb_resultset_create(void)
 
 void neb_cb_resultset_destroy(neb_cb_resultset *cbrs)
 {
+	if(cbrs == NULL)
+		return;
 	g_ptr_array_free(cbrs->cb_results, TRUE);
 	nm_free(cbrs);
 }


tvestelind:
I wouldn't mind having a unit test for this but I'm fine with merging this.

tvestelind:
I added a unit test for this and it's currently on review internally.

tvestelind:
If you would like to see and review it then I can either create a new PR (if I understand GH's documentation correctly) or you can look at my commit here: https://github.com/tvestelind/naemon-core/commit/cf2b6b3dc79f01aec1e6402a842f2d4736634fb2

sni:
Great, looks good to me. Afaik you could even push the test case onto this PR branch.

tvestelind:
Nice, then I'll consider it ready to merge.

It says `Add more commits by pushing to the fix_segfault_without_eventbroker branch on sni/naemon-core.` but I don't think it I have access to push to your fork, or am I missing something?

sni:
> but I don't think it I have access to push to your fork, or am I missing something?

Thats a checkbox when creating a pull request. So yes, you should have permission to push to that branch in my repo. But i don't care, we can also do 2 PRs if thats easier :-)

tvestelind:
Ah ok, good to know for the future. I'll try to remember to do the same :)

PR #173, opened 2017-04-04
fix comment ids increment on reload

sni:
Comments related to downtimes are beeing added with persistance=false flag. This
leads to the comment beeing removed during a reload. The downtime then immediatly
adds the comment again but the id increments each reload. Furter this results in
a warning from livestatus for each downtime on every reload:
livestatus: Cannot delete non-existing downtime/comment

### DIFF
--- a/src/naemon/downtime.c
+++ b/src/naemon/downtime.c
@@ -486,7 +486,7 @@ int register_downtime(int type, unsigned long downtime_id)
 
 
 	/* add a non-persistent comment to the host or service regarding the scheduled outage */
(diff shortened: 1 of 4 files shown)


sni:
I don't think so, otherwise we would have noticed this earlier :-) I am not aware of any unit tests which cover reloading the core, are you?

tvestelind:
Nope. I know that there are some tests but I'm not sure what they do and how they do it.

sni:
I added a test case and fixed a issue with the previous patch. Should be fine now(tm)

sni:
Any more comments?

PR #175, opened 2017-04-12
Feature/mon 10198 fail naemon start when daemonizing

roengstrom:
Doing configuration changes in a large environments with big configuration files and then restarting naemon will take a very long time. We would like to remove all verification done when starting and restarting and make configuration verification explicit through `naemon -v`.

This patch removes the verification done in the init script. It also makes the parent wait for the child to be fully initialized before exiting the parent when daemonizing, making it possible to catch if the child fails during the initialization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ Makefile
 *.gcov
 *.a
 *.so
+*.pyc
 gmon.out
 core.[1-9]*
 /Documentation
--- a/features/daemonize.feature
+++ b/features/daemonize.feature
@@ -1,0 +1,26 @@
+Feature: Daemonization
+    Naemon should successfully daemonize if and only if there is a valid
+    configuration available.
+
+    Scenario: Naemon fails to daemonize when an invalid system config is
+        provided
+        Given I have an invalid naemon system configuration
+        And config verification fail
+        Then naemon should fail to start
+
+    Scenario: Naemon fails to daemonize when an invalid object config is
+        provided
+        Given I have an invalid naemon object configuration
+        And config verification fail
+        Then naemon should fail to start
+
+    Scenario: Naemon fails to daemonize when all config provided is invalid
+        Given I have an invalid naemon object configuration
+        And I have an invalid naemon system configuration
+        And config verification fail
+        Then naemon should fail to start
+
+    Scenario: Naemon successfully daemonizes when a valid config is provided
+        Given config verification pass
+        Then naemon should successfully start
+
--- a/features/environment.py
+++ b/features/environment.py
@@ -1,0 +1,47 @@
+from support.naemon_object_config import NaemonObjectConfig
+from support.naemon_system_config import NaemonSystemConfig
+import tempfile
+import os.path
+import signal
+import shutil
+
+
+def before_all(context):
+    os.environ['LD_LIBRARY_PATH'] = (
+        os.getcwd() + '/' + context.config.userdata['shared_libs_path']
+    )
+
+
+def before_scenario(context, scenario):
+    context.wrkdir = os.getcwd()
+    context.tmpdir = tempfile.mkdtemp()
+    assert os.path.exists(context.tmpdir) is not False, (
+        'Failed to create a temporary directory'
+    )
+    print ('Changing directory to %s from %s' % (
+            context.tmpdir, context.wrkdir
+        )
+    )
+    os.chdir(context.tmpdir)
+    context.naemonsysconfig = NaemonSystemConfig()
+    context.naemonobjconfig = NaemonObjectConfig()
+
+
+def after_scenario(context, scenario):
+    # Kill any naemon daemonsv
+    if os.path.isfile('naemon.pid'):
+        pid = int(open('naemon.pid').read())
+        try:
+            os.kill(pid, signal.SIGTERM)
+            print ('Killed the naemon process (%i)' % pid)
+        except OSError as e:
+            print (os.strerror(e.errno))
+            pass
+    print ('Changing directory to %s from %s' % (
+            context.wrkdir, context.tmpdir
+        )
+    )
+    os.chdir(context.wrkdir)
+    if scenario.status != 'failed':
+        print ('Deleting temporary directory %s' % context.tmpdir)
+        shutil.rmtree(context.tmpdir, True)
--- a/features/steps/naemon.py
+++ b/features/steps/naemon.py
@@ -1,0 +1,89 @@
+from behave import given, when, then, use_step_matcher
+import subprocess
+
+use_step_matcher('re')
+
+
+@given('I have naemon (?P<object_type>.+) objects')
+def add_naemon_object(context, object_type):
+    print ("Object type %s" % object_type)
+    for row in context.table:
+        context.naemonobjconfig.add_obj(object_type, context.table)
+
+
+@given('I have naemon config (?P<parameter>.+) set to (?P<value>.+)')
+def set_naemon_parameter(context, parameter, value):
+    context.naemonsysconfig.set_var(parameter, value)
+
+
+@given('I have an invalid naemon system configuration')
+def invalid_sysconfig(context):
+    context.execute_steps(u'Given I have naemon config invalid_param set to x')
+
+
+@given('I have an invalid naemon object configuration')
+def invalid_objconfig(context):
+    context.execute_steps(u'''
+        Given I have naemon host objects
+            | use          | host_name    | address   | check_command |
+            | default-host | invalid_host | 127.0.0.1 | non_existing  |
+    ''')
+
+
+@given('I write config to file')
+def configuration_to_file(context):
+    context.naemonobjconfig.to_file(context.naemonobjconfig.filename)
+    context.naemonsysconfig.to_file(context.naemonsysconfig.filename)
+    context.execute_steps(u'Given I have directory checkresults')
+
+
+@given('I verify the naemon configuration')
+def config_verification(context):
+    context.execute_steps(u'Given I write config to file')
+    executable = (
+        context.wrkdir + '/' + context.config.userdata['naemon_exec_path']
+    )
+    args = [executable, '--allow-root', '-v', context.naemonsysconfig.filename]
+    context.return_code = subprocess.call(args)
+
+
+@given('config verification fail')
+def config_verification_fail(context):
+    context.execute_steps(u'Given I verify the naemon configuration')
+    assert context.return_code != 0, (
+        'Return code was %s' % context.return_code
+    )
+
+
+@given('config verification pass')
+def config_verification_pass(context):
+    context.execute_steps(u'Given I verify the naemon configuration')
+    assert context.return_code == 0, (
+        'Return code was not 0 (got %s)' % context.return_code
+    )
+
+
+@when('I start naemon')
+def naemon_start(context):
+    context.execute_steps(u'Given I write config to file')
+    executable = (
+        context.wrkdir + '/' + context.config.userdata['naemon_exec_path']
+    )
+    args = [executable, '--allow-root', '-d', context.naemonsysconfig.filename]
+    context.return_code = subprocess.call(args)
+
+
+@then('naemon should fail to start')
+def naemon_start_fail(context):
+    context.execute_steps(u'When I start naemon')
+    assert context.return_code != 0, (
+        'Return code was %s' % context.return_code
+    )
+
+
+@then('naemon should successfully start')
+def naemon_start_success(context):
+    context.execute_steps(u'When I start naemon')
+    assert context.return_code == 0, (
+        'Return code was not 0 (got %s)' % context.return_code
+    )
--- a/features/support/naemon_object_config.py
+++ b/features/support/naemon_object_config.py
@@ -1,0 +1,102 @@
+class NaemonObjectConfig(object):
+
+    filename = 'oconf.cfg'
+
+    def __init__(self):
+        self.current_config = {
+            'contact': [{
+                'name': 'default-contact',
+                'contact_name': 'default-contact',
+                'host_notifications_enabled': '1',
+                'service_notifications_enabled': '1',
+                'host_notification_period': '24x7',
+                'service_notification_period': '24x7',
+                'host_notification_options': 'a',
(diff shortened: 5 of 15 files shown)


tvestelind:
Several behave tests fail in travis but that is not reflected here. That must be fixed.

roengstrom:
Fixed the Travis build. Now all tests are running and are passing.

features/steps/naemon.py
```diff
@@ -0,0 +1,75 @@
...
+        context.wrkdir + '/' + context.config.userdata['naemon_exec_path']
+    )
+    args = [executable, '--allow-root', '-v', context.naemonsysconfig.filename]
+    context.return_code = subprocess.call(args)
+
+
+@given('config verification fail')
+def config_verification_fail(context):
+    context.execute_steps(u'Given I verify the naemon configuration')
+    assert context.return_code is not 0, (
```
tvestelind: Use `==`

Read (a bit) more here:  http://stackoverflow.com/a/2239753/4965690

features/steps/naemon.py
```diff
@@ -0,0 +1,75 @@
...
+        'Return code was %s' % context.return_code
+    )
+
+
+@given('config verification pass')
+def config_verification_pass(context):
+    context.execute_steps(u'Given I verify the naemon configuration')
+    assert context.return_code is 0, (
```
tvestelind: `==`

features/steps/naemon.py
```diff
@@ -0,0 +1,75 @@
...
+        context.wrkdir + '/' + context.config.userdata['naemon_exec_path']
+    )
+    args = [executable, '--allow-root', '-d', context.naemonsysconfig.filename]
+    context.return_code = subprocess.call(args)
+
+
+@then('naemon should fail to start')
+def naemon_start_fail(context):
+    context.execute_steps(u'When I start naemon')
+    assert context.return_code is not 0, (
```
tvestelind: `==`

features/steps/naemon.py
```diff
@@ -0,0 +1,75 @@
...
+        'Return code was %s' % context.return_code
+    )
+
+
+@then('naemon should successfully start')
+def naemon_start_success(context):
+    context.execute_steps(u'When I start naemon')
+    assert context.return_code is 0, (
```
tvestelind: `==`

features/support/naemon_object_config.py
```diff
@@ -0,0 +1,139 @@
...
+        for type, objects in self.current_config.items():
+            for object in objects:
+                res += ('define %s {\n' % (type))
+                for key, value in object.items():
+                    res += ('\t%s\t%s\n' % (key, value))
+                res += ('}\n\n')
+        return res
+
+    def to_file(self, filename):
+        f = open(filename, 'w+')
```
tvestelind: Why `w+` here and not just `w`?

features/support/naemon_object_config.py
```diff
@@ -0,0 +1,139 @@
...
+                    ' servicestate="$SERVICESTATE$"'
+                    ' serviceoutput="$SERVICEOUTPUT$"'
+                    ' longserviceoutput="$LONGSERVICEOUTPUT$"'
+                )
+            }, {
+                'command_name': 'check_long_output',
+                'command_line': './output_cmd'
+            }, {
+                'command_name': 'check_active_ok',
+                'command_line': './active_ok_cmd'
```
tvestelind: same as `exit 0` right?

features/support/naemon_object_config.py
```diff
@@ -0,0 +1,139 @@
...
+            }, {
+                'command_name': 'check_active_critical',
+                'command_line': './active_critical_cmd'
```
tvestelind: `exit 2`

features/support/naemon_object_config.py
```diff
@@ -0,0 +1,139 @@
...
+                    ' hostname="$HOSTNAME$"'
+                    ' servicedesc="$SERVICEDESC$"'
+                    ' notificationcomment="$NOTIFICATIONCOMMENT$"'
+                    ' notificationtype="$NOTIFICATIONTYPE$"'
+                    ' servicestate="$SERVICESTATE$"'
+                    ' serviceoutput="$SERVICEOUTPUT$"'
+                    ' longserviceoutput="$LONGSERVICEOUTPUT$"'
+                )
+            }, {
+                'command_name': 'check_long_output',
```
tvestelind: Do we need this command?

features/support/naemon_object_config.py
```diff
@@ -0,0 +1,139 @@
...
+                'command_line': './output_cmd'
+            }, {
+                'command_name': 'check_active_ok',
+                'command_line': './active_ok_cmd'
+            }, {
+                'command_name': 'check_active_critical',
+                'command_line': './active_critical_cmd'
+            }],
+            'host': [{
+                'check_command': 'log_check_host',
```
tvestelind: Could just be `check_active_ok` to simplify things

features/support/naemon_object_config.py
```diff
@@ -0,0 +1,139 @@
...
+                'notification_options': 'a',
+                'notifications_enabled': 1,
+                'register': 0,
+                'name': 'default-host',
+            }],
+            'service': [{
+                'is_volatile': 0,
+                'max_check_attempts': 3,
+                'check_interval': 5,
+                'check_command': 'log_check_service',
```
tvestelind: Could just be `check_active_ok` to simplify things

features/support/naemon_object_config.py
```diff
@@ -0,0 +1,139 @@
...
+                'alias': '24x7',
+                'sunday': '00:00-24:00',
+                'monday': '00:00-24:00',
+                'tuesday': '00:00-24:00',
+                'wednesday': '00:00-24:00',
+                'thursday': '00:00-24:00',
+                'friday': '00:00-24:00',
+                'saturday': '00:00-24:00'
+            }],
+            'command': [{
```
tvestelind: Most of these commands can be removed to slim down the config.

features/support/naemon_system_config.py
```diff
@@ -0,0 +1,34 @@
...
+        self.current_config[name] = value
+
+    def to_string(self):
+        res = ''
+        for key, value in self.current_config.items():
+            res += ("%s=%s\n" % (key, value))
+        return res
+
+    def to_file(self, filename):
+        f = open(filename, 'w+')
```
tvestelind: Same here about `w+`, I think `w` should be enough right?

src/naemon/utils.c
```diff
@@ -535,14 +545,62 @@ int daemon_init(void)
...
+			nm_log(NSLOG_RUNTIME_ERROR, "Unable to read from pipe: %s",
+				strerror(errno));
+			returnCode = EXIT_FAILURE;
+		}
+		if (close(upipe_fd[PIPE_READ]) < 0) {
+			nm_log(NSLOG_RUNTIME_ERROR, "Unable to close parent read end: %s",
+				strerror(errno));
+			returnCode = EXIT_FAILURE;
+		}
+		if (returnCode < 0) {
```
tvestelind: Should this be `returnCode != OK` ?

features/environment.py
```diff
@@ -0,0 +1,41 @@
...
+    print ('Changing directory to %s from %s' % (
+            context.tmpdir, context.wrkdir
+        )
+    )
+    os.chdir(context.tmpdir)
+    context.naemonsysconfig = NaemonSystemConfig()
+    context.naemonobjconfig = NaemonObjectConfig()
+
+
+def after_scenario(context, scenario):
```
roengstrom: Sure, even though I have a feeling it could be useful during development to be able to read logs even after a successful run, I guess we shouldn't clutter peoples systems by default when running tests. Will fix.

features/steps/naemon.py
roengstrom: Did you mean `!=`?

src/naemon/utils.c
roengstrom: I don't think it really changes anything, but it does make things more explicit and maybe more clear. Will fix.

features/steps/naemon.py
tvestelind: Ah yes :) The same for the other `is not` further down.

features/support/naemon_object_config.py
roengstrom: No, at this time this is dead code, so will remove it.

features/startup.feature
```diff
@@ -0,0 +1,29 @@
+Feature: Startup
```
roengstrom: But it does say in the Feature description `Naemon should start up if and only if there is a valid configuration`. Do you think that is lacking?

tvestelind: It's a little bit ambiguous as we're only testing with the `-d` flag and not without it.

.gitignore
```diff
@@ -26,6 +26,8 @@ Makefile
 *.gcov
 *.a
 *.so
+*.pyc
+*.json
```
tvestelind: We can remove this now as we're no longer outputting to behave.json

roengstrom:
@tvestelind So I've taken all your comments into consideration and adjusted the code accordingly.

features/steps/naemon.py
```diff
@@ -0,0 +1,89 @@
...
+    context.execute_steps(u'Given I have directory checkresults')
+
+
+@given('I verify the naemon configuration')
+def config_verification(context):
+    context.execute_steps(u'Given I write config to file')
+    executable = (
+        context.wrkdir + '/' + context.config.userdata['naemon_exec_path']
+    )
+    args = [executable, '--allow-root', '-v', context.naemonsysconfig.filename]
```
tvestelind: Could we use `--verify-config` here instead of `-v`?

features/steps/naemon.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+
+
+@when('I start naemon')
+def naemon_start(context):
+    context.execute_steps(u'Given I write config to file')
+    executable = (
+        context.wrkdir + '/' + context.config.userdata['naemon_exec_path']
+    )
+    args = [executable, '--allow-root', '-d', context.naemonsysconfig.filename]
```
tvestelind: Can we use `--daemon` here instead of `-d` ?

src/naemon/utils.c
```diff
@@ -535,14 +545,62 @@ int daemon_init(void)
...
+			nm_log(NSLOG_RUNTIME_ERROR, "Unable to read from pipe: %s",
+				strerror(errno));
+			returnCode = EXIT_FAILURE;
+		}
+		if (close(upipe_fd[PIPE_READ]) < 0) {
+			nm_log(NSLOG_RUNTIME_ERROR, "Unable to close parent read end: %s",
+				strerror(errno));
+			returnCode = EXIT_FAILURE;
+		}
+		if (returnCode != 0) {
```
tvestelind: I think that we should compare with either `OK` or `EXIT_FAILURE` as those are the values that are assigned to `returnCode`

src/naemon/utils.c
```diff
@@ -535,14 +545,62 @@ int daemon_init(void)
...
+	}
 
-	/* child continues... */
+	if ((pid = (int)fork()) < 0) {
+		nm_log(NSLOG_RUNTIME_ERROR, "Unable to fork out the daemon process: %s",
+			strerror(errno));
+		return (ERROR);
+	} else if (pid != 0) {
+		/* parent stops - when child is done, we exit here */
+		int returnCode = OK;
```
tvestelind: I think `return_code` would follow the style better

features/daemonize.feature
```diff
@@ -0,0 +1,25 @@
...
+    Naemon should successfully daemonize if and only if there is a valid
+    configuration available.
+
+    Scenario: Naemon fails to daemonize when an invalid system config is
+        provided
+        Given I have an invalid naemon system configuration
+        And config verification fail
+        Then naemon should fail to start
+
+    Scenario: Naemon fails to daemonize when an invalid object config is
```
swedieman: This scenario is identical to the scenario above. Why do we need this?

features/steps/naemon.py
roengstrom: Not without Flake8 complaining, the width of the row would exceed 79 and we'd have to introduce line breaks or make it into a variable with a shorter name. Personally I don't think it's worth it, do you?

features/steps/naemon.py
roengstrom: Same case as with the --verify-config flag.

src/naemon/utils.c
roengstrom: Agree!

features/daemonize.feature
```diff
@@ -0,0 +1,25 @@
...
+
+    Scenario: Naemon fails to daemonize when all config provided is invalid
+        Given I have an invalid naemon object configuration
+        And I have an invalid naemon system configuration
+        And config verification fail
+        Then naemon should fail to start
+
+    Scenario: Naemon successfully daemonizes when a valid config is provided
+        Given config verification pass
+        Then naemon should successfully start
```
swedieman: Detail, but look at the sign at the end of this line - file should end with a newline

src/naemon/utils.c
roengstrom: Good point. Just pushed some changes.

features/daemonize.feature
roengstrom: In each scenario we're testing two different types of configuration (system config and object config). We want to make sure naemon fails to initialize in both cases.

features/steps/naemon.py
tvestelind: I do :) But it's not that important.

features/daemonize.feature
swedieman: Ah, it wasn't identical! I blame my eyes

tvestelind: :)

features/daemonize.feature
roengstrom: All done! 👍

tvestelind:
Good, then we can merge it.

Could you push it internally and we'll merge it.

To clarify for anyone else reading this. We'll try to review more on Github in the future but for now we must merge in-house. We're working on changing that as well so that we can merge here directly in the future.

PR #176, opened 2017-04-20
reduce log spam

sni:
Do not log timeouts 3 times for each timeout. Timeouts are a normal thing in
monitoring and state changes are logged already, there is no need to add 2
extra log entries, one from wproc and one from the check handler. Especially since
the timeout is quite visible already and the extra logging does not add any value.

### DIFF
--- a/src/naemon/checks_host.c
+++ b/src/naemon/checks_host.c
@@ -408,8 +408,6 @@ int update_host_state_post_check(struct host *hst, struct check_result *cr)
 	/* adjust return code (active checks only) */
 	if (cr->check_type == CHECK_TYPE_ACTIVE) {
 		if (cr->early_timeout) {
-			nm_log(NSLOG_RUNTIME_WARNING,
-			       "Warning: Check of host '%s' timed out after %.2lf seconds\n", hst->name, hst->execution_time);
 			nm_free(hst->plugin_output);
 			nm_free(hst->long_plugin_output);
 			nm_free(hst->perf_data);
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -437,8 +437,6 @@ int handle_async_service_check_result(service *temp_service, check_result *queue
 	nm_free(temp_service->perf_data);
 
 	if (queued_check_result->early_timeout == TRUE) {
-		nm_log(NSLOG_RUNTIME_WARNING,
-		       "Warning: Check of service '%s' on host '%s' timed out after %.3fs!\n", temp_service->description, temp_service->host_name, temp_service->execution_time);
 		nm_asprintf(&temp_service->plugin_output, "(Service check timed out after %.2lf seconds)", temp_service->execution_time);
 		temp_service->current_state = service_check_timeout_state;
 	}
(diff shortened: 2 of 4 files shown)


src/naemon/checks_host.c
```diff
@@ -408,8 +408,8 @@ int update_host_state_post_check(struct host *hst, struct check_result *cr)
 	/* adjust return code (active checks only) */
 	if (cr->check_type == CHECK_TYPE_ACTIVE) {
 		if (cr->early_timeout) {
-			nm_log(NSLOG_RUNTIME_WARNING,
-			       "Warning: Check of host '%s' timed out after %.2lf seconds\n", hst->name, hst->execution_time);
+			//nm_log(NSLOG_RUNTIME_WARNING,
```
mattiasr: Please remove code instead of hide behind comment tag

src/naemon/checks_service.c
```diff
@@ -437,8 +437,8 @@ int handle_async_service_check_result(service *temp_service, check_result *queue
 	nm_free(temp_service->perf_data);
 
 	if (queued_check_result->early_timeout == TRUE) {
-		nm_log(NSLOG_RUNTIME_WARNING,
-		       "Warning: Check of service '%s' on host '%s' timed out after %.3fs!\n", temp_service->description, temp_service->host_name, temp_service->execution_time);
+		//nm_log(NSLOG_RUNTIME_WARNING,
```
mattiasr: Please remove code instead of hide behind comment tag

src/worker/worker.c
```diff
@@ -270,7 +270,7 @@ static void kill_job(struct nm_event_execution_properties *event)
 	 */
 	if (cp->ei->state != ESTALE) {
 		timeouts++;
-		wlog("Killing job %d with pid %d due to timeout. timeouts=%u; started=%u", id, pid, timeouts, started);
+		//wlog("Killing job %d with pid %d due to timeout. timeouts=%u; started=%u", id, pid, timeouts, started);
```
mattiasr: Please remove code instead of hide behind comment tag

sni:
You are right, i changed that. Thanks for the fast review :-)

src/worker/worker.c
```diff
@@ -270,7 +270,6 @@ static void kill_job(struct nm_event_execution_properties *event)
 	 */
 	if (cp->ei->state != ESTALE) {
 		timeouts++;
```
tvestelind: `id` and `pid` are no longer used. Either remove the assignments or refactor to use them where we use `cp->ei->pid`. Which ever is fine by me.

sni:
alright :-) i remove them as well.

tvestelind:
Can you rebase this to the latest commit on master? I think that we might have introduced a bug in the last PR so we reverted it.

roengstrom:
@tvestelind We're going to re-revert it and also add a fix for the introduced bug in a few minutes.

sni:
rebased...

tvestelind:
From tests:
```
tests/test-worker.c:99:F:worker reaping tests:worker_test_timeout:0: Expected 'due to timeout' to be present in log

FAIL: tests/test-worker
```

sni:
oh man, these tests are a nightmare... We have `t/` and `tests/` and `tap/` and even a `t-tap/`, it feels like everyone created his own test folder.

tvestelind:
Yea, it's not very clear what each folder tests either :/

sni:
Ok, i moved most wproc: message to the debug logfile and adjusted the tests.

rhagman:
Looks good to me, well done!

PR #177, opened 2017-04-21
Bugfix/daemon child dies when worker dies

roengstrom:


roengstrom:
This is a re-revert of the pull request merged a few days ago followed with a fix for the bug it introduced.

src/naemon/utils.c
```diff
@@ -556,23 +555,13 @@ int daemon_init(void)
...
-	}
-
 	if ((pid = (int)fork()) < 0) {
 		nm_log(NSLOG_RUNTIME_ERROR, "Unable to fork out the daemon process: %s",
 			strerror(errno));
 		return (ERROR);
 	} else if (pid != 0) {
 		/* parent stops - when child is done, we exit here */
-		int return_code = OK;
+		int return_code = EXIT_FAILURE;
```
tvestelind: Could we just have a comment where this variable is set to `OK`, just to make things a bit clearer.

roengstrom: Absolutely, fixed!

tvestelind:
I'm OK with merging internally :+1:

PR #179, opened 2017-04-27
naemon executable path

roengstrom:
Added ability to omit setting the executable path to the
naemon binary and the shared libraries. Also added possibility
to use absolute paths.

Signed-off-by: Robin Engström <robin.engstrom@op5.com>

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -7,9 +7,16 @@ import shutil
 
 
 def before_all(context):
-    os.environ['LD_LIBRARY_PATH'] = (
-        os.getcwd() + '/' + context.config.userdata['shared_libs_path']
-    )
+    # check if we have set relative paths in behave.ini
+    if 'shared_libs_path' in context.config.userdata:
+        os.environ['LD_LIBRARY_PATH'] = \
+            os.path.abspath(context.config.userdata['shared_libs_path'])
+
+    # use naemon executable defined in PATH if nothing else is given
+    context.naemon_exec_path = 'naemon'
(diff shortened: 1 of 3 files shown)


mattiasr:
Looks good to me

features/environment.py
```diff
@@ -7,9 +7,22 @@
 
 
 def before_all(context):
-    os.environ['LD_LIBRARY_PATH'] = (
-        os.getcwd() + '/' + context.config.userdata['shared_libs_path']
-    )
+    # check if we have set relative paths in behave.ini
+    if 'shared_libs_path' in context.config.userdata:
+        shared_libs_path = ''
+        if context.config.userdata['shared_libs_path'][0] == '/':
```
tvestelind: Wouldn't `os.path.abspath(context.config.userdata['shared_libs_path'])` achieve the same thing as all this?

features/environment.py
```diff
@@ -7,9 +7,22 @@
...
+            shared_libs_path = context.config.userdata['shared_libs_path']
+        else:
+            shared_libs_path = os.getcwd() + '/' + \
+                context.config.userdata['shared_libs_path']
+        os.environ['LD_LIBRARY_PATH'] = shared_libs_path
+
+    if 'naemon_exec_path' in context.config.userdata:
+        if context.config.userdata['naemon_exec_path'][0] != '/':
```
tvestelind: Same here for abspath

features/environment.py
roengstrom: That's way nicer, thanks! Updated code pushed.

PR #180, opened 2017-05-05
undeprecate check_result_path

sni:
Reading check results from disk is not a bad thing. Lot of addons use this for good reasons.
Ex. Thruk submits business process results by spool folder. We could make check_result_path
optional, but i don't recommend removing it. Therefor we remove the deprecation warning.

### DIFF
--- a/src/naemon/configuration.c
+++ b/src/naemon/configuration.c
@@ -181,8 +181,6 @@ read_config_file(const char *main_config_file, nagios_macros *mac)
 		}
 
 		else if (!strcmp(variable, "check_result_path")) {
-			obsoleted_warning(variable, "Support for processing check results from disk will be removed");
-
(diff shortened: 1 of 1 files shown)


tvestelind:
Github seems to be a bit buggy for me so I cannot make comments in the file. But looking quickly there might be issues related to these features. This one is from the commit that deprecated that feature https://jira.op5.com/browse/MON-8381. I'll try and look more to see why we decided to no longer support it.

sni:
I don't think the mentionen issue is a problem. It is even solved already.

tvestelind:
From my perspective this is merge:able. Why do you say @roengstrom and @mattiasr ?

mattiasr:
Sure this can be merged, we don't use it so for our point of view it's safe.

roengstrom:
I agree, don't have any problem with this being merged.

sni:
Thanks

PR #182, opened 2017-05-09
daemon: return faster when daemonizing

tvestelind:
Return after all the APIs that naemon supply are ready, i.e. the query handler

This is of course up for discussion. Do you feel that something else needs to be initialized before we consider the start complete?

Signed-off-by: Tomas Vestelind <tomas.vestelind@op5.com>

### DIFF
--- a/src/naemon/naemon.c
+++ b/src/naemon/naemon.c
@@ -529,6 +529,15 @@ int main(int argc, char **argv)
 		nerd_init();
 		timing_point("NERD initialized\n");
 
+		/* let the parent know we're good to go and that it can let go */
+		if (daemon_mode == TRUE && sigrestart == FALSE) {
+			if ((result = signal_parent(OK)) != OK) {
+				broker_program_state(NEBTYPE_PROCESS_SHUTDOWN, NEBFLAG_PROCESS_INITIATED, NEBATTR_SHUTDOWN_ABNORMAL);
+				cleanup();
+				exit(ERROR);
+			}
+		}
+
 		/* initialize check workers */
 		if (init_workers(num_check_workers) < 0) {
 			nm_log(NSLOG_RUNTIME_ERROR, "Failed to spawn workers. Aborting\n");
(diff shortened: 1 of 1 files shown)


sni:
This means when NEB modules fail to load, the parent won't notice.

roengstrom:
So this would equal the behaviour we had earlier when daemonizing, that we say we've started OK if the configuration looks okay and the fork succeeds. So I'd be okay with this behaviour.

src/naemon/naemon.c
```diff
@@ -529,6 +529,15 @@ int main(int argc, char **argv)
 		nerd_init();
 		timing_point("NERD initialized\n");
 
+		/* let the parent know we're good to go and that it can let go */
```
roengstrom: Shouldn't we signal the parent after the child has checked the object configuration? I.e. after `result = read_all_object_data(config_file);`

tvestelind:
I'm looking at other projects and did some tests:
# Load non-existing module 
# Load a random file of the same type as a module 
# Pass an invalid argument to the proper module

Results:
* Apache
## Returns 1 with print out about the module not existing
## Returns 1 with print out about the module not implementing the specified API
## Returns 0, crashes shortly after with errors in error-log
* nginx
## Returns 1 with print out about the module not existing
## Returns 1 with print out about the module not implementing the specified API
## Returns 1 with print with argument not being correct

So, I think we can go either way. Both of the software that I've looked at at least load the modules and verify that their API is correct. I think the easiest here would to load and init the modules before saying that we've started correctly.

sni:
What are we trying to achieve here? We just made the parent wait till everything has started and now we start excluding some things from that. We should either rollback the previous way and just fire/forget the daemon or accept that loading modules takes some time. We don't make anything
faster if the child signals the parent earlier, its just like before then when the child said "yes, it might work, but we cannot be sure yet, just assume everything is ok"

tvestelind:
You are right. The reason we wanted to return earlier is to speed up restarting a whole cluster using merlin. There are things that we do in the start process that takes considerable time but maybe we should work on solving that (or remove them if they are not necessary) instead of returning early. 
Thank you for your feedback. I will close this PR.

Issue #174, opened 2017-04-06
external command CHANGE_SVC_MODATTR does not work


anfoe1111:
When issuing the external command CHANGE_SVC_MODATTR to reset the "modified attributes" of a service we observe the following entry in /var/log/naemon/naemon.log:
 **[timestamp] Warning: External command parse error [timestamp] CHANGE_SVC_MODATTR;_our_hostname_;_our_servicename_;0 (Failed validation of service as type service (argument 1))**
This happens when issuing the command manually (printf to commandfile) and also when using thruk.
We tested all kind of services: no chance to reset the "modified attributes".
Any ideas? We run naemon 1.0.6 stable and thruk 2.14 on top of centos 7.
Regards,
Andreas

anfoe1111:
Further investigations:

When changing the following line in commands.c everything works fine:
original code:
...
core_command = command_create("CHANGE_SVC_MODATTR", service_command_handler,
			"This command changes the modified attributes value for the specified service. Modified attributes values are used by Naemon to determine which object properties should be retained across program restarts. Thus, modifying the value of the attributes can affect data retention. This is an advanced option and should only be used by people who are intimately familiar with the data retention logic in Naemon.", "host=host_name;service=service;ulong=value");
...

changed to:  ("host=host_name;" removed)
...
core_command = command_create("CHANGE_SVC_MODATTR", service_command_handler,
			"This command changes the modified attributes value for the specified service. Modified attributes values are used by Naemon to determine which object properties should be retained across program restarts. Thus, modifying the value of the attributes can affect data retention. This is an advanced option and should only be used by people who are intimately familiar with the data retention logic in Naemon.", "service=service;ulong=value");
...

Could this probably be a bug?
Regards,
Andreas

sni:
You are right, there is a "host=host_name" too much. Thanks for pointing this out. Could you create a Pull Request for this?

PR #185, opened 2017-06-01
external command CHANGE_SVC_MODATTR does not work

anfoe1111:
see issue #174


sni:
Looks good, thanks.

tvestelind:
If it looks good to @sni then I'm fine with this. I'll create a PR in-house.

Issue #186, opened 2017-06-08
CHANGE_SVC_MODATTR still not working


anfoe1111:
see issue #174 , still not working when merging pull-request

Finally I found a second error in commands.c (probably a copy&paste typo), tests are running fine now, I'll send a pull-request.

Regards,
Andreas

anfoe1111:
corresponding pull request #187 was signed off, everything ok

PR #187, opened 2017-06-08
fix external command CHANGE_SVC_MODATTR

anfoe1111:
fix for issues #174 and #186

### DIFF
--- a/src/naemon/commands.c
+++ b/src/naemon/commands.c
@@ -2397,7 +2397,7 @@ static int service_command_handler(const struct external_command *ext_command, t
 
 		case CMD_CHANGE_SVC_MODATTR:
 			target_service->modified_attributes = GV_ULONG("value");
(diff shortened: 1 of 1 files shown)


sni:
I guess you are right :-)

roengstrom:
Looks okay to me. Will run this PR internally and see if we can merge it.

rhagman:
Would be nice to be able to test this so that we don't break it in the future.

tsjondin:
Nice catch, though a test for the command is required for merge.

sni:
The fact that we don't have test cases for all commands should not prevent this PR from being merged, especially if its just a simple typo fixed.

tvestelind:
It's not just a typo, the functionality is changed.

I don't think it's unreasonable to demand a test for this, unless the test is very hard to write. That being said, we should help to check whether we can write such a test or not.

tsjondin:
I believe a simple test, creating a host, a service and then running this command against the service is sufficient. Not all commands have to be tested in this test, just the functionality that has changed.

tvestelind:
If making a test is hard then I can go with @sni's suggestion and merge it

Issue #184, opened 2017-05-31
Can not compile on GCC 7.1.1


modulatix:
GCC 7 introduced format-truncation and implicit-fallthrough
With default -Werror flag naemon-core does not compile as such warnings yielded by code
https://pastebin.ca/3822910

PR #188, opened 2017-06-12
Support for GCC 7

modulatix:
Done via preprocessor macros so users of pre-7 GCC will notice no changes

### DIFF
--- a/src/naemon/checks.c
+++ b/src/naemon/checks.c
@@ -207,6 +207,9 @@ int process_check_result_queue(char *dirname)
 	char *temp_buffer = NULL;
 	int result = OK, check_result_files = 0;
 	time_t start;
+#if __GNUC__ >= 7
+	unsigned int writesize;
+#endif
 
 	/* make sure we have what we need */
 	if (dirname == NULL) {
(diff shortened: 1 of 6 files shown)


src/naemon/checks.c
```diff
@@ -240,9 +243,16 @@ int process_check_result_queue(char *dirname)
 		}
 
 		/* create /path/to/file */
+                
+#if __GNUC__ >= 7
+		writesize = snprintf(file, sizeof(file), "%s/%s", dirname, dirfile->d_name);
+		if (writesize < (strlen(dirname) + strlen(dirfile->d_name) + 2)) 
+			nm_log(NSLOG_RUNTIME_WARNING, "Warning: truncated path to file in check result queue directory '%s'", dirfile->d_name);
```
modulatix: Not sure if this a right way to such things, as this can create tons of warnings. 
Also onelining would make this unreadable, so I've added 'int writesize'

tvestelind:
Thank you for your PR! I'm sorry but I think using the CPP to {in,ex}clude code like this make a code base very messy.

What are the GCC-errors that you are trying to solve? Is it the one about format truncation?

modulatix:
Format-truncation and switch-case fall-through
Format-truncation fix could be made global without preprocessor magic (just checking return value of sprintf is fine AFAIK), so GCC7 will not complain, but fall-through warning can be only suppressed by attribute or -Wimplicit-fallthrough=0. Or disabling -Wall

tvestelind:
@modulatix nice summary. I think that checking the return value should be the way to go. Maybe they could even be rewritten to use the `nm_asprinf` function.
As for the implicit fall through. I.m.o. we should go with either using `-Wimplicit-fallthrough=0` or rewriting the code but not disabling `-Wall`. A re-write would be the nicest but that could mean a lot of work..

modulatix:
@tvestelind i can try to do this hard way without preprocessor. I think i can do this before monday

tvestelind:
That sounds great! 👍

tvestelind:
If you need help please let me know and I'll see what I can do :)

jacobbaungard:
Hi,

I have added an alternative solution at #217 . I started working on the issue (#184) before I noticed there was already a pull request up. The alternate solution uses comments instead of GCC attributes for fall through, and relies on the follow property of snprintf: 

> Thus, a return value of size or more means that the output was truncated
https://linux.die.net/man/3/snprintf

roengstrom:
A fix for this without preprocessor was merged with PR #217

PR #190, opened 2017-06-16
Documentation fix for Sticky Acknowledgements

eythori-sensa:


tvestelind:
Thank you for the PR! Is this (http://www.naemon.org/documentation/developer/externalcommands/acknowledge_svc_problem.html) also wrong then? If so, how can we update it @sni ?

sni:
There is a manual update script here:
https://github.com/naemon/naemon.github.io/blob/master/documentation/developer/externalcommands/update
Which need to be run after the PR got merged.

PR #191, opened 2017-06-19
Fixes memory leak for DATE/TIME Macros

dirtyren:
The DATE / TIME macros were not freeing the allocated pointer.
 I had to swap mkstr for asprintf in order to be able to free the pointer outside the function.

### DIFF
--- a/src/naemon/macros.c
+++ b/src/naemon/macros.c
@@ -584,18 +584,19 @@ int grab_datetime_macro_r(nagios_macros *mac, int macro_type, char *arg1, char *
 		break;
 
 	case MACRO_TIMET:
-		*output = (char *)mkstr("%lu", (unsigned long)current_time);
+		asprintf(&*output, "%lu", (unsigned long)current_time);
 		break;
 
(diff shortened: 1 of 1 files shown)


src/naemon/macros.c
```diff
@@ -584,18 +584,18 @@ int grab_datetime_macro_r(nagios_macros *mac, int macro_type, char *arg1, char *
 		break;
 
 	case MACRO_TIMET:
-		*output = (char *)mkstr("%lu", (unsigned long)current_time);
+		asprintf(&*output, "%lu", (unsigned long)current_time);
```
tvestelind: Couldn't this be just `output` instead of `&*output`?

sni:
Also please use the `nm_asprintf` macro which does automatically return code checking.

src/naemon/macros.c
dirtyren: Now that you mentioned it... it probably can. :)

dirtyren:
Done, changes made.

PR #194, opened 2017-07-06
remove shadownaemon

sni:
shadownaemon was meant to shadow a live copy of a real naemon process over livestatus. This has several unsolvable flaws, like timeperiods not working, etc...
meanwhile there is the LMD project which solves most of the issues, so deprecate shadownaemon here and save some code.


tvestelind:
👏 👍

PR #195, opened 2017-07-06
fix newline escaping for check results

sni:
parse_check_outputs() escape_newlines_please argument has been ignored but it
is explicitly requested from the worker.  This leads to issues in csv output of
livestatus, since livestatus assumes the longout put to be escaped while it was
not. So it returned newlines just as is instead of escaped ones.

### DIFF
--- a/src/naemon/checks.c
+++ b/src/naemon/checks.c
@@ -186,8 +186,13 @@ int parse_check_output(char *buf, char **short_output, char **long_output, char 
 	struct check_output *check_output = nm_malloc(sizeof(struct check_output));
 	check_output = parse_output(buf, check_output);
 	*short_output = check_output->short_output;
-	*long_output = check_output->long_output;
 	*perf_data = check_output->perf_data;
+	if(escape_newlines_please == TRUE && check_output->long_output != NULL) {
+		*long_output = g_strescape(check_output->long_output, "");
+		free(check_output->long_output);
+	} else {
+		*long_output = check_output->long_output;
+	}
 	free(check_output);
 	strip(*short_output);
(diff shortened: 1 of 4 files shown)


tvestelind:
Looks good and really nice with a test for it! 

Now there's only the issue with the two failing tests.

sni:
Oh thanks for the pointer, didn't notice the retention tests failing :-) Will fix them...

sni:
Alright

sni:
Any thoughts on this?

tvestelind:
Copying from our internal code review, I guess @roengstrom forgot to add it here:
```
The documentation for g_strescape() says

    Escapes the special characters '\b', '\f', '\n', '\r', '\t', '\v', '' and '"' in the string source by inserting a '' before them.

Should we really escape everything if escape_newlines_please is TRUE? Intuitively It seems like we should only escape newlines.
```

Sorry :/

sni:
Maybe we can rename that function later on, for now it doesn't matter as long as we use 
g_strcompress() to reverse escaping. Its also only used to safely store the state into the retention file. So i'd say it doesn't matter here.

roengstrom:
I think we could rename `escape_newlines_please` with something like `escape_output`. But if you @sni and @tvestelind feel like it's fine as it is I'm okay with merging. :)

tvestelind:
I'm fine with whatever 👍

sni:
i am fine with that as well, we can/should refactor `parse_check_output` later. For example the last argument `newlines_are_escaped` isn't used at all anywhere in the code.

roengstrom:
Nice work @sni

PR #201, opened 2017-09-08
Make systemd startup cleaner

nycaleksey:
Prevent naemon systemd service from complaining about existing /var/run/naemon directory.


sni:
Thanks, looks good to me.

PR #202, opened 2017-09-09
cleanup memory leak

jframeau:
to make valgrind output a bit cleaner.


sni:
This patch does have some side effects at least. Retrieving the check_source via livestatus doesn't work anymore when applying this patch.

sni:
This has been fixed in Mod-Gearman. The check_source is explicitly a "non-freeable" const char* and Mod-Gearman used a allocated char*.

PR #206, opened 2017-09-19
decouple core, livestatus and thruk

sni:
Having packaging files for livestatus and core in one folder is a bad idea
because each component has different release cycles. After moving thruk into
its own packages, this is the next consequent step and decouples core and
livestatus.
This allows us to build naemon-core packages without having to rebuild livestatus.
It also makes space for a clean naemon meta package.

While cleaning up, the tools subpackage has been included in core now since
it only contains one single small binary only.

In order to make rpmbuild work, the op5build folder has to go, but vendor
specific files don't belong here anyway.

This PR also add 2 new make targets:

 - make rpm
 - make deb

which will build the corresponding packages.


sni:
Any chance to have a look at this?

PR #207, opened 2017-09-20
Revert "Merge pull request #73 in MONITOR/naemon from bugfix/fix_newl…

roengstrom:
…ine_escaping to master"

After the merge of this patch other naemon modules started failing, such as Merlin.
We shouldn't adapt the naemon api after another module. But instead adapt
the module in question according to the api.

This reverts commit c722025af4e6d588cd9af784f340117703dcafc0, reversing
changes made to 9ac4d20155ac805bd26849915b21192f328e5bf0.

Signed-off-by: Robin Engström <robin.engstrom@op5.com>

### DIFF
--- a/src/naemon/checks.c
+++ b/src/naemon/checks.c
@@ -186,13 +186,8 @@ int parse_check_output(char *buf, char **short_output, char **long_output, char 
 	struct check_output *check_output = nm_malloc(sizeof(struct check_output));
 	check_output = parse_output(buf, check_output);
 	*short_output = check_output->short_output;
+	*long_output = check_output->long_output;
 	*perf_data = check_output->perf_data;
-	if(escape_newlines_please == TRUE && check_output->long_output != NULL) {
-		*long_output = g_strescape(check_output->long_output, "");
-		free(check_output->long_output);
-	} else {
-		*long_output = check_output->long_output;
-	}
 	free(check_output);
 	strip(*short_output);
 	strip(*perf_data);
--- a/src/naemon/xrddefault.c
+++ b/src/naemon/xrddefault.c
@@ -157,6 +157,7 @@ int xrddefault_save_state_information(void)
 	/* save host state information */
 	for (temp_host = host_list; temp_host != NULL; temp_host = temp_host->next) {
 		struct host *conf_host;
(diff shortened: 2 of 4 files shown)


rhagman:
Looks good to me!

sni:
I'd assume other modules expect naemon to behave closely like nagios which we reverted here again. I'd say Merlin should be fixed and soon and then we should fix this again.

sni:
Oh, and the comment id bugfix got reverted as well? Is this on purpose?

roengstrom:
I'm fine with reverting for now, fixing merlin as soon as possible and then apply this patch again. Did I understand you correctly?

Don't see the comment id bugfix being included here, am I missing it?

sni:
You are right, the comment id commit was mentioned in the commit text.

And also right, please fix Merlin and the reapply the changes. I really would like to do a release soon with all those fixes.

roengstrom:
Oh yeah, sorry for the confusion. Thanks, will fix Merlin asap.

nook24:
Is there some documentation about API changes available?
At the moment module maintainers always need to track all commits done to Naemon, or just try and error...

But as long as `g_strescape` will not break json, I'm fine with this :)

sni:
`g_strescape` won't break json, but it will change how long plugin output is stored internally (and served by livestatus). With that change, newlines will be exported as newlines, just like with the original mk-livestatus. Before that change you had to unescape the literal '\n' by yourself.

nook24:
Ok good, many thanks for the information @sni

sni:
Any news on this? Were you able to address this issue in merlin?

roengstrom:
Have been able to work around the patch but it became a lot harder to test. So we haven't been able to write a fix with tests so nothing is merged. Will have another look at the tests for Merlin today and tomorrow so hopefully we can remerge the patch if that's ok?

By the way, did you verify the long output in notifications with your patch?

sni:
> By the way, did you verify the long output in notifications with your patch?

I will do that...

roengstrom:
Sorry that this is taking time. We spent a couple of days during our last sprint trying to fix our tests but it has shown to be a complicated matter. We're up to our necks in urgent issues that we're working on this week and it doesn't look like we're going to be able to spend any more time on it this week. :(

Issue #200, opened 2017-09-08
Memory leak or heavy memory usage


jframeau:
Centos 7 (7.3.1611)
OMD 2.40 - naemon 1.0.6.

The facts:

- 2000 hosts
- 9000 active services
- mod_gearman and livestatus modules.

naemon starts at 25 Mo and is 1 Go of RAM after 12h (and continue to grow unless a cold restart, reload doesn't help).

**I found a first leak using valgrind:**

In mod_gearman (neb_module/result_thread.c, line 229), there's a strdup which naemon doesn't release.

#ifdef USENAEMON
chk_result->source = gm_strdup( value );
#endif

So my proposition in naemon (head):

--- a/src/naemon/checks.c
+++ b/src/naemon/checks.c
@@ -524,6 +524,7 @@ int free_check_result(check_result *info)
nm_free(info->host_name);
nm_free(info->service_description);
nm_free(info->output);
nm_free(info->source);

return OK;
}

Second, more obscur for me, using valgrind tool massif. After 30 mn, one code seems to use heavely memory:

->62.33% (**64,728,793B**) 0x4E9347E: nm_bufferqueue_push (bufferqueue.c:283)
| ->59.13% (61,401,535B) 0x4E6CDA6: update_service_performance_data (perfdata.c:499)
| | ->59.13% (61,401,535B) 0x4E45209: handle_async_service_check_result (checks_service.c:994)
| |   ->59.13% (61,401,535B) 0x82BE547: ???
| |     ->59.13% (61,401,535B) 0x4E56FF3: execute_and_destroy_event.constprop.6 (events.c:249)
| |       ->59.13% (61,401,535B) 0x4E5757F: event_poll (events.c:367)
| |         ->59.13% (61,401,535B) 0x4032B3: main (in /opt/omd/versions/2.40-labs-edition/bin/naemon)

This is not a memory leak.

Could one explain why that code has so huge memory usage ?

jfr

jframeau:
Full valgrind report using massif tool for information:

[memory.txt](https://github.com/naemon/naemon-core/files/1288148/memory.txt)

dirtyren:
You could download the latest build  to test if the memory leak is still so massive.
 This pull request was accepted that fixed a memory leak https://github.com/naemon/naemon-core/pull/191

  I still have a memory leak happening in the latest build that I could not locate yet. Maybe with the patch your valgrid log will show less leaks.

  []s.

sni:
[...]

PR #210, opened 2017-09-26
fix memory leak when collecting performance data

sni:
We need to check the service_perfdata_fd/host_perfdata_fd as well when pushing
performance data onto our buffer. Checking the template does not work because
it is never NULL as it will be set to a default if its unset.

This should fix issue #200.

### DIFF
--- a/src/naemon/perfdata.c
+++ b/src/naemon/perfdata.c
@@ -483,6 +483,9 @@ static int xpddefault_update_service_performance_data_file(nagios_macros *mac, s
 	if (svc == NULL)
 		return ERROR;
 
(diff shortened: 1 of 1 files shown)


tvestelind:
Would it be possible to return faster from `update_{service,host}_performance_data` instead?

sni:
I thought about that as well, but i guess we need to check the filehandle in `xpddefault_update_service|host_performance_data_file` anyway and then we would have the same if clause twice with probably almost no benefit.
Also this is out of the scope of this bugfix, this one just plugs a memory leak :-)

PR #211, opened 2017-10-04
Naemon starts SIGKILLing the wrong processes if the PID wrap-around is too short

sidhartha-sankar:
MON-10495
Naemon starts SIGKILLing the wrong processes if the PID wrap-around is too short

### DIFF
--- a/src/worker/worker.c
+++ b/src/worker/worker.c
@@ -249,23 +249,27 @@ static int get_process_parent_id(const pid_t pid, pid_t * ppid) {
 
         sprintf(buffer, "/proc/%d/stat", pid);
         fp = fopen(buffer, "r");
-        if (!fp)
+        if (!fp) {
                 return errno;
+        }
 
         fread(buffer, sizeof (char), sizeof (buffer), fp);
         errreading = ferror(fp);
-        if (fclose(fp) != 0)
+        if (fclose(fp) != 0) {
                 return errno;
+        }
 
-        if (errreading != 0)
+        if (errreading != 0) {
                 return errreading;
+        }
 
         strtok(buffer, " "); // (1) pid  %d
         strtok(NULL, " "); // (2) comm  %s
         strtok(NULL, " "); // (3) state  %c
 
-        if ( (s_ppid = strtok(NULL, " ")) == NULL) // (4) ppid  %d
+        if ( (s_ppid = strtok(NULL, " ")) == NULL) {// (4) ppid  %d
                 return EBADF;
+        }
 
         *ppid = atoi(s_ppid);
         return 0;
--- a/src/worker/worker.c
+++ b/src/worker/worker.c
@@ -244,7 +244,7 @@ static int finish_job(child_process *cp, int reason)
  */
 static int get_process_parent_id(const pid_t pid, pid_t * ppid) {
         char buffer[BUFSIZ], *s_ppid;
-        int errreading;
+        int errreading, size;
         FILE *fp;
 
         sprintf(buffer, "/proc/%d/stat", pid);
@@ -253,7 +253,7 @@ static int get_process_parent_id(const pid_t pid, pid_t * ppid) {
                 return errno;
         }
 
-        fread(buffer, sizeof (char), sizeof (buffer), fp);
+        size = fread(buffer, sizeof (char), sizeof (buffer), fp);
         errreading = ferror(fp);
         if (fclose(fp) != 0) {
                 return errno;
--- a/src/worker/worker.c
+++ b/src/worker/worker.c
@@ -1,3 +1,6 @@
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
 #include "naemon/events.h"
 #include "worker.h"
 #include "lib/worker.h"
@@ -237,6 +240,27 @@ static int finish_job(child_process *cp, int reason)
 }
 
 /*
+ * Get the parent PID from a PID
+ */
+static void get_process_parent_id(const pid_t pid, pid_t * ppid) {
+	char buffer[BUFSIZ], *s_ppid;
+	FILE *fp;
+	sprintf(buffer, "/proc/%d/stat", pid);
+	fp = fopen(buffer, "r");
+	if (fp) {
+		size_t size = fread(buffer, sizeof (char), sizeof (buffer), fp);
+		if (size > 0) {
+			strtok(buffer, " "); // (1) pid  %d
+			strtok(NULL, " "); // (2) comm  %s
+			strtok(NULL, " "); // (3) state  %c
+			s_ppid = strtok(NULL, " "); // (4) ppid  %d
+			*ppid = atoi(s_ppid);
+		}
+		fclose(fp);
+	}
+}
+
+/*
  * "What can the harvest hope for, if not for the care
  * of the Reaper Man?"
  *   -- Terry Pratchett, Reaper Man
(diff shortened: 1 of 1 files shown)


src/worker/worker.c
```diff
@@ -263,6 +288,14 @@ static void kill_job(struct nm_event_execution_properties *event)
 		(void)kill(-cp->ei->pid, SIGKILL);
 		return;
 	}
+	/* check if the child we'r killing belongs to this worker process */
+	wpid = getpid();
+	get_process_parent_id(pid, &ppid);
+	if (-1 != ppid && ppid != wpid) {
```
roengstrom: If we can't find the process, we proceed to sending a SIGKILL further down below?
I think this should be `if (ppid == -1 || ppid != wpid)`.

sidhartha-sankar: I think it's the os (operating system ) issue if we can't find a running process in os proc list.
Another case (process just finished and we din't find in proc list).
In my opinion, though it's rare case but in both cases going down may not be harmful.

roengstrom: A third case is where a process spawns right after we check for its existence. Given it's a slim chance, but it could happen if we're spawning a whole lot of processes, which we are in a large naemon system. Then we'll again be sending SIGKILL to a process that isn't the workers child.

sni:
Thats what i mentioned already. There are probably a lot of corner cases were we will still kill the wrong process. I think its better to have 10 processes still running than killing 1 wrong one.
Also, less code, less complicated.

sidhartha-sankar:
@sni Currently (existing naemon code) we are killing so many wrong processes.
I think this fix would restrict that to a large extend.

src/worker/worker.c
sidhartha-sankar: Then I think it's ok to modify this `if ... case` .

roengstrom:
My opinion is that we need this fail safe as a self preservation mechanism. I also believe a lot of users depend on having a timeout for their plugins. But I don't think we should be sending SIGKILL to a PID which isn't in the process list upon check. Since some users may be running several hundred checks per second the PID count will be increasing very fast and the PID may in fact exist upon sending the SIGKILL after the PID check.

sni:
Sure, we should definitly kill the direct child of a worker (the check plugin) when hitting the timeout.

src/worker/worker.c
```diff
@@ -263,6 +288,14 @@ static void kill_job(struct nm_event_execution_properties *event)
 		(void)kill(-cp->ei->pid, SIGKILL);
 		return;
 	}
+	/* check if the child we'r killing belongs to this worker process */
+	wpid = getpid();
+	get_process_parent_id(pid, &ppid);
+	if (ppid == -1 || ppid != wpid) {
```
ricardomaraschini: Now if the function fails this evaluates to `true`. Ah, by the way: if `ppid == -1`  it will always be `!= wpid`. Am I wrong?

src/worker/worker.c
```diff
@@ -237,6 +240,27 @@ static int finish_job(child_process *cp, int reason)
...
+		if (size > 0) {
+			strtok(buffer, " "); // (1) pid  %d
+			strtok(NULL, " "); // (2) comm  %s
+			strtok(NULL, " "); // (3) state  %c
+			s_ppid = strtok(NULL, " "); // (4) ppid  %d
+			*ppid = atoi(s_ppid);
+		}
+		fclose(fp);
+	}
+}
```
ricardomaraschini: I refactored this one, adding a return value and checking in almost all places where something can go wrong. What do you think?
```c
static int get_process_parent_id(const pid_t pid, pid_t * ppid) {
        char buffer[BUFSIZ], *s_ppid;
        int errreading;
        FILE *fp;

        sprintf(buffer, "/proc/%d/stat", pid);
        fp = fopen(buffer, "r");
        if (!fp)
                return errno;

        size_t size = fread(buffer, sizeof (char), sizeof (buffer), fp);
        errreading = ferror(fp);
        if (fclose(fp) != 0)
                return errno;

        if (errreading != 0)
                return errreading;

        if (size == 0)
                return EBADF;

        strtok(buffer, " "); // (1) pid  %d
        strtok(NULL, " "); // (2) comm  %s
        strtok(NULL, " "); // (3) state  %c

        // (4) ppid  %d
        if ( (s_ppid = strtok(NULL, " ")) == NULL)
                return EBADF;

        *ppid = atoi(s_ppid);
        return 0;
}
```

sidhartha-sankar: Any error checking for this line `*ppid = atoi(s_ppid);` ?

sidhartha-sankar: I think checking `fclose(fp)` after checking `if (!fp)` isn't necessary. We'r checking a lot of os(operating system) system calls in this function.

ricardo-maraschini: Regarding the `atoi` call: that is exactly why I commented that i validated *almost* all error possibilities, I would use `strtonum` instead. Now, related to your last comment, yeah, there are *a lot* of places where things can go wrong, to be sure or not that everything went well depends on how faithful you are.

sidhartha-sankar: Ok. Would like to know if anyone else has any more comments on this function code?

src/worker/worker.c
roengstrom: Yes, if we fail to find the pid we were looking for this will evaluate to `true` so we can return early. I think you're right with your last statement so perhaps the first evaluation is unnecessary. Although I'm fine with the way it is now. Makes it somewhat easier to read.

src/worker/worker.c
sidhartha-sankar: Done!

src/worker/worker.c
```diff
@@ -237,6 +240,38 @@ static int finish_job(child_process *cp, int reason)
...
+ * Get the parent PID from a PID
+ */
+static int get_process_parent_id(const pid_t pid, pid_t * ppid) {
+        char buffer[BUFSIZ], *s_ppid;
+        int errreading;
+        FILE *fp;
+
+        sprintf(buffer, "/proc/%d/stat", pid);
+        fp = fopen(buffer, "r");
+        if (!fp)
```
tvestelind: Use `{ }` please.

tvestelind: This applies to all other places too of course

src/worker/worker.c
```diff
@@ -237,6 +240,38 @@ static int finish_job(child_process *cp, int reason)
...
 /*
+ * Get the parent PID from a PID
+ */
+static int get_process_parent_id(const pid_t pid, pid_t * ppid) {
+        char buffer[BUFSIZ], *s_ppid;
+        int errreading;
+        FILE *fp;
+
+        sprintf(buffer, "/proc/%d/stat", pid);
+        fp = fopen(buffer, "r");
```
tvestelind: Is it not possible to use `open` or what's the reason for using `fopen`?

sidhartha-sankar: Yes we can use `open` instead of `fopen`. But I think it's better to use this library function than this system call.

src/worker/worker.c
sidhartha-sankar: Updated. Thnx.

src/worker/worker.c
tvestelind: Why is it better?

sidhartha-sankar: I found a nice reason here: https://stackoverflow.com/questions/1658476/c-fopen-vs-open

sidhartha-sankar: I found a nice reason here: https://stackoverflow.com/questions/1658476/c-fopen-vs-open

tvestelind: Thanks, I learned something new :)

tvestelind:
Looks good to me 👍

PR #214, opened 2017-12-07
Perform nudging to reschedule at 1 second next interval

felixmarch:
When the max_concurrent_checks is set, currently the naemon just simply drops the check and reschedules it to next interval when the retry_interval is not defined (src/naemon/checks_service.c:145)

As a result, if the next interval is 1 minute, the next check will be scheduled to next 1 minute slot, causing the system to be idling for at most 1 minute.

If we have 200 service checks, then the delay will be very long, up to 200 minutes. With this, critical service check will not be able to run immediately when the system is back to idle and has to lose 1 minute interval

So, we propose to make use of "nudge_seconds" variable to perform reschedule in one second interval attempt.

Easy test would be to create 200 service checks and set max_concurrent_checks to 1 second. With max_concurrent_checks 1 second, the 200 service checks has to be able to run 200 checks in serial mode without making the system idle for 1 minute during that serial execution

### DIFF
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -144,12 +144,8 @@ static void handle_service_check_event(struct nm_event_execution_properties *evp
 					nm_log(NSLOG_RUNTIME_WARNING,
 					       "\tMax concurrent service checks (%d) has been reached.  Nudging %s:%s by %d seconds...\n", max_parallel_service_checks, temp_service->host_name, temp_service->description, nudge_seconds);
 				} 
-				/* Simply reschedule at retry_interval instead, if defined (otherwise keep scheduling at next 1 second) */
-				if (temp_service->retry_interval != 0.0) {
-					schedule_next_service_check(temp_service, get_service_retry_interval_s(temp_service), 0);
-				} else {
-					schedule_next_service_check(temp_service, nudge_seconds, 0);
-				}
+				/* Simply reschedule at next 1 second */
+				schedule_next_service_check(temp_service, nudge_seconds, 0);
 				return;
 			} else {
 				if (nudging_in_progress == 1) { /* Print the last log when it has been recovered */
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -113,7 +113,7 @@ void schedule_service_check(service *svc, time_t check_time, int options)
 static void handle_service_check_event(struct nm_event_execution_properties *evprop)
 {
 	service *temp_service = (service *)evprop->user_data;
-	int nudge_seconds = 0;
+	int nudge_seconds = 1;
 	double latency;
 	struct timeval tv;
 	struct timeval event_runtime;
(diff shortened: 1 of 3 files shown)


sni:
I like it, till now, `max_concurrent_checks` was quite useless. I haven't looked deeper into your changes yet, but i would lower the loglevel to debug.

ageric:
It might be a while since I looked at this, but max_concurrent_checks doesn't track seconds (and never has). It's simply the max number of checks that can be run at any given time. If you set it to one, you're effectively configuring your system to only ever run one check at any given time. If any of your checks then take a long time to complete, that will cause long latencies for all other checks in your system.

You should almost always leave max_concurrent_checks at 0, which is taken to mean "spawn checks as required to meet the scheduling requirements".

ageric:
On a secondary note, adding runtime tracking variables for "engine state" to objects breaks the ABI, and so requires a major release, as all modules (livestatus, merlin, etc) need to be updated and/or recompiled in order to not break. A compatible way of doing it would be to introduce a data segment located hash table (ie, one declared "static" in the file where it's used) which tracks if the check is being nudged or not.

nook24:
Could you please document the required changes for NEB modules? It's very hard for module developers that there is no change log and documentation. We need to track all source code changes, to bring our modules back to live.

felixmarch:
Hello @ageric . Thanks for the highlight and comment. 

I am not clear on what you mean on introducing a data segment located hash table.

Is your highlight about the new variable nudging_in_progress in the objects_service.h? 

If so, would it be fine to move it to checks_service.c like this?

```
diff --git a/src/naemon/checks_service.c b/src/naemon/checks_service.c
index 41d0a28..0ce417b 100644
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -40,6 +40,8 @@ static void check_for_orphaned_services_eventhandler(struct nm_event_execution_p
 /* Status functions, immutable */
 static int is_service_result_fresh(service *, time_t, int);

+/* nudging status */
+static int nudging_in_progress = 0;

 /******************************************************************************
  *******************************  INIT METHODS  *******************************
diff --git a/src/naemon/objects_service.c b/src/naemon/objects_service.c
index f9c396f..d15d66d 100644
--- a/src/naemon/objects_service.c
+++ b/src/naemon/objects_service.c
@@ -13,7 +13,6 @@
 #include "lib/libnaemon.h"

 static GHashTable *service_hash_table;
-int nudging_in_progress = 0;
 service *service_list = NULL;
 service **service_ary = NULL;

diff --git a/src/naemon/objects_service.h b/src/naemon/objects_service.h
index 1159f05..3e33280 100644
--- a/src/naemon/objects_service.h
+++ b/src/naemon/objects_service.h
@@ -19,7 +19,6 @@ typedef struct service service;
 struct servicemember;
 typedef struct servicesmember servicesmember;

-extern int nudging_in_progress;
 extern struct service *service_list;
 extern struct service **service_ary;
```

The reason we set max_concurrent_checks not to 0, is because we are getting high load average issue caused by the concurrent checks. 

Your colleague Jonatan Sundeen found limiting the max_concurrent_checks to 10 could resolve the high load average issue. 

However, after applying max_concurrent_checks, we noticed new issue that many critical checks were missing due to this long latency issue. 

So, we'd like to have this max_concurrent_checks issue to be fixed as soon as possible because we rely on this max_concurrent_checks to prevent high load average in our environment.

Thanks.

ageric:
What I mean is that adding variables to the object structures causes modules to no longer work, so adding a variable to the service object means we have to bump the object ABI version, which is troublesome. Also, I don't quite see the point behind this patch. Checks should be automagically nudged if the max_concurrent_checks limit is reached anyway. If they're not, then it should be fixed globally, and not on a per-service level.

That aside though; If you're getting high load average issues with max_concurrent_checks set to 0, your hardware is underdimensioned for the network you want to monitor and the frequency with which you want to run the checks. Try increasing check_interval instead.

You should also be aware that a system with 16 cpu's (for example) can safely have a high load average and still not be running at high capacity.

My thinking is therefore that this patch has come about as a fix to a bad solution for a different problem. If I'm correct, that means that fixing the original problem the correct way makes this patch unneeded.

felixmarch:
Hello @ageric 
How will you propose to fix it globally, and not on a per-service level? Does adding static variable in checks_service.c also need to bump the object ABI version?

We hope we can fix it on our side, but currently not able to. There are some critical check our customers need to perform and do not allow us to increase check_interval. Moreover, the script is just simple check (using generic snmpwalk/snmpget command), running over 100+ devices. We are not able to control hardware/infrastructure where network can sometimes slow down the check.

sni:
For the time beeing, you could have a look at Mod-Gearman (https://mod-gearman.org/) addon which makes it possible to serially execute checks either globally or on a servicegroup basis.

ageric:
The most likely culprit for the high load is inefficient plugins though. It's one of the few drawbacks of the Naemon/Nagios way of running checks, that resourcehungry plugins can make the entire machine sluggish. SNMP plugins written in perl are notorious for this, so if that's what you're using you should probably switch to the ones found in https://github.com/ageric/plugins-op5-check_snmp instead.

It's an SNMP shim written in C and a bunch of plugins coded around it that I wrote while I was working at op5, simply because all script languages we tried were absolutely horrible with regards to performance.

If you're using plugins that read and write files as part of their execution (like checking traffic on switches etc) you should also make sure you disable the 'modify on read' flag for the mounted disk. That will cut your disk write accesses essentially in half.

So swapping from inefficient plugins to efficient ones would be step one. Once that's done, setting max_concurrent_checks back to 0 (ie, unlimited) should be doable. If it's not, we should start looking into modifying the core rescheduling logic.

You must understand that keeping the check schedule when the machine is underdimensioned for it is an absolute impossibility though. You just can't run 1,000,000 real checks per minute on a raspberry pi for instance, so if max_concurrent_checks=0 isn't working for you now, it's not gonna work in the future with just minor patches to the core. Me and Sven (@sni) both put a *lot* of effort into maximizing the efficiency of the core a few years back, so I'm pretty sure that whatever's hogging resources on your system lies somewhere else, and nudging checks one second at a time isn't going to solve your problem.

felixmarch:
Thank you @ageric and @sni

Will try your plugins-op5-check_snmp and mod_gearman from @sni

It would be good if op5 can include this mod_gearman as well for its customers 

Last week I did try to install mod_gearman but getting "unrecognized service" when starting the service. Was having some time to figure it out and learn similar issue reported at the forum (https://groups.google.com/forum/#!topic/mod_gearman/rueEiWpK8e8)

Eventually I realized I needed to run "yum install gearmand-server".

ageric:
Please let us know how it pans out. It will also help whoever's looking to solve similar problems in the future

jacobbaungard:
I wonder if we should close this one?

PR #215, opened 2018-02-10
do not escape double quotes

sni:
they would end up escaped in the pluginoutput and break html output for example.


sni:
ah, the original #195 has been reverted. This patch will apply if #195 gets merged again.

sni:
already merged.

PR #217, opened 2018-03-16
Allow compilation with GCC 7

### LINKED ISSUE #184: Can not compile on GCC 7.1.1

jacobbaungard:
GCC 7 introduced new warnings which, as we compile Naemon with -Werror, meant that the builds failed.

The warnings are trigged when building naemon are:

-Wimplicit-fallthrough which warns if there are any fallthroughs in switch/case statements.

-Wformat-truncation which warns if output from a formatting operating could be truncated (for example when running snprintf)

These two patches ensures that Naemon can be correctly built with GCC 7. See the individual commit messages for further information.

This should fix #184

### DIFF
--- a/lib/runcmd.c
+++ b/lib/runcmd.c
@@ -199,6 +199,7 @@ int runcmd_cmd2strv(const char *str, int *out_argc, char **out_argv)
 				set_state(STATE_INSQ | STATE_INARG);
 				continue;
 			}
+			/* FALLTHROUGH */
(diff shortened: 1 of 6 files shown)


roengstrom:
Looks good to me

PR #218, opened 2018-04-09
Triggered Scheduled downtime

tnsetting:
The triggered scheduled downtime is deleted when the triggering downtime is starting.
This issue is due to that flex downtime start variable is feeded with right value. It will add a wrong time (negative) for the downtime stop event on the event queue and the downtime stop event for the triggered downtime will start evecution right after it is started.
The solution is to initialize the flex downtime start variable of the triggered downtime with the value of the flex downtime start value of the triggering one.


jacobbaungard:
This has been merged into master with: 2a4d59422739159c3dabfb6908b5dc576f616dcb and I am therefore closing this PR.

PR #219, opened 2018-04-09
Notification is sent even if a service is in a scheduled flexible downtime

tnsetting:
When a service is set in flexible downtime, the first notification is sent out before the downtime schedule takes effect.
Since service result check logic is just add the downtime start event to the event queue and then start notification right after,
this bug is natural with this flow.
The solution is checking whether there is any flexible downtime started and if not we will send out the normal notifciation.


jacobbaungard:
This has been merged into master with: 2a4d59422739159c3dabfb6908b5dc576f616dcb and I am therefore closing this PR.

PR #221, opened 2018-04-17
Logging for switching service states for host down 

tnsetting:
Adding a log message to explain why the hard states switching happened.

### DIFF
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -794,8 +794,10 @@ int handle_async_service_check_result(service *temp_service, check_result *queue
 			log_debug_info(DEBUGL_CHECKS, 2, "Host is not UP, so we mark state changes if appropriate\n");
 
 			/* "fake" a hard state change for the service - well, its not really fake, but it didn't get caught earlier... */
-			if (temp_service->last_hard_state != temp_service->current_state)
+			if (temp_service->last_hard_state != temp_service->current_state) {
 				hard_state_change = TRUE;
+				nm_log(NSLOG_INFO_MESSAGE, "SERVICE INFO: %s;%s; Service switch to hard down state due to host down.\n", temp_service->host_name, temp_service->description);
+			}
 
 			/* update last state change times */
 			if (state_change == TRUE || hard_state_change == TRUE)


jacobbaungard:
As I understand it, this change will add a log message for every single service which is moved to a hard down state due to the host being down.  This might spam the log quite a bit if a host have many services.

Is it possible to add something to the existing service alert if it is moved to a hard down state instead? I.e something like:

`[2016-04-05 14:49:01] SERVICE ALERT: host1;svc1;CRITICAL;HARD;1;CRITICAL - Hard state switch due to host down - Socket timeout after 10 seconds`

I am however not sure if the above format, will cause problems for other tools which parses the service alerts?

tnsetting:
Currently the logging logic for service alert is a bit generic function which is printing states switching basically. The last field for the alert is the output from the check plugin.  I am not sure it is a good idea to alter this information.

I think it will span the log a bit but this scenario does not happen that regularly.

jacobbaungard:
Okay, if changing the current output log-line is infeasible, I am OK with adding a new log line as is done here.

I am however a little bit concerned that this does not have the same format as the other SERVICE ALERT. That might be problematic for tools parsing the logs.

I think we should probably take one of these two approaches:

1) Follow the existing service alert format ( check objects_service.c ) - and add a "NOTE: .." as the message or something similar. But then perhaps if something parses the logs, it will see two service alerts with the same status (expect for the message), not sure if that will be confusing/cause problems.

2) Do not use the service alert format at all, but add a log entry that does not contain "SERVICE ALERT". Maybe it could be "SERVICE NOTE" or something similar. The disadvantage of this, is if there are tools the parses the logs and look for events related to a specific service, then this message might not show up. Those tools could be updated to support this new log entry though.

I am would be leaning towards approach 2). Any additional comments is welcome though @roengstrom .

tnsetting:
Currently the log type is set to NSLOG_INFO_MESSAGE instead other service related type NSLOG_SERVICE*. I agree with the second option that the message heading should be something like SERVICE NOTE or SERVICE INFO since it is an additional message to the SERVICE ALERT message and it will not confuse the existing parsing tools.

PR #225, opened 2018-05-09
Double naemon process

tnsetting:
Double naemon process situtation is created if other process is monitoring naemon process and try to restart it.
Naemon relies on the pid file locking to ensure other instance of naemon could not be started. But the advisory locking breaks if the
pid file is remove. Bascially you can always create a new naemon process if you just remove the pidfile.
In the commit, the pidfile is not removed after naemon process exiting the main function and the modified the stop function in the
init.d file so that the pidfile is not removed after killing the naemon process.

### DIFF
--- a/daemon-init.in
+++ b/daemon-init.in
@@ -96,7 +96,9 @@ stop() {
     elif type start-stop-daemon >/dev/null 2>&1; then
         start-stop-daemon --stop --name ${prog} --pidfile ${pidfile} --exec ${exec}
     else
-        killproc -p ${pidfile} ${exec}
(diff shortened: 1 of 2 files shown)


tnsetting:
No really. 
At start up, naemon opens the pidfile. If not exist, it will create the pidfile.
Then naemon check if anyone is locking the pidfile. If someone locking, it will exit immediately. 
Fork.
Naemon tries to lock the pidfile using advisory locking. If locking ok, naemon continues. If not, naemon exits. This is the place no other naemon can lock the pidfile.

The problem. If the pidfile is moved, removed or renamed, you can start another naemon instance and the existing naemon will still be running. And you can repeat the process and create multiple naemon instance.

PR #226, opened 2018-05-15
first notification delay should start on hard state

sni:
first_notification_delay should start on the last hard change instead of using
the last time ok. Hard state changes could also occur if a service changes from
warning to critical.

Example with max_check_attempts 5
 soft 0: OK
 soft 1: Warning
 soft 2: Warning
 soft 3: Warning
 soft 4: Critical
 hard 5: Critical

Naemon would now use the time from the last ok (soft 0) to calculate the first
notification delay instead of the actual time of the usual first notification (hard 5)

references:
  - https://github.com/NagiosEnterprises/nagioscore/pull/483
  - https://github.com/NagiosEnterprises/nagioscore/pull/482


sni:
already merged.

PR #230, opened 2018-05-25
Prepare next release

sni:
this pull request contains everything required for the next release except the release itself.


jacobbaungard:
Is it possible to change the commit author on "Reload : Fix defunct" so it is not "root" ?

sni:
sure, i faked the author and rebased the commits.

sni:
The commit also changes the location in the debian and rpm package. I could amend those changes to the initial pkg commit or reword the commit description if you like.

jacobbaungard:
It is not really a major issue, but I feel it would be cleaner to have them in the initial pkg commit instead.

sni:
Ok, i moved the spec and debian/ changes to the first commit and rebased everything.

t/900-configparsing.t
```diff
@@ -15,6 +15,7 @@ my $naemon = "$buildroot/src/naemon/naemon";
 my $etc = "$buildroot/t/etc";
 my $precache = "$buildroot/t/var/objects.precache";
 
+plan skip_all => 'config check won\' run as root user' if $> == 0;
```
jacobbaungard: Rather than skipping the test we could instead pass the --allow-root argument to the Naemon binary.

sni:
yeah, i thought that as well but i was too lazy :-) its fixed now

PR #231, opened 2018-05-28
Remove op5build

jacobbaungard:
Vendor specific files should be in a downstream repository instead.


sni:
yeah, thanks.

PR #232, opened 2018-05-31
Added changelog for 1.0.7 release

jacobbaungard:
An attempt to create a change log for the 1.0.7 release.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,1 +1,1 @@
-See `git log` and its various options and alternatives.
+NEWS
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,49 @@
+1.0.7 - Jun 01 2018
+===================
+Features:
+  * Allow circular dependencies in Naemons host graph (experimental)
+
+Changed:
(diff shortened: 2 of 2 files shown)


jacobbaungard:
I moved the NEWS file to the CHANGELOG and removed the ChangeLog file from before. I am not sure whether this is the best practice though. Maybe it is better to keep it in the NEWS file. Not sure.

Or perhaps we should follow something like: https://keepachangelog.com/en/1.0.0/ instead.

I left out some commits, such as those that only touches testing (for example https://github.com/naemon/naemon-core/commit/8720d6db7e8f5aa8cc1bd9ce7d18bbf2f0d696db), or things that did minor code cleanup (for example: https://github.com/naemon/naemon-core/commit/0335087529552c9c4706cb3f75b9b873779dc65f).

Hopefully everything else should be included.

sni:
The changelog itself looks good, thanks for your work. The news and changelog file must be changed in the spec file as well and in the Makefile.am.

PR #234, opened 2018-06-01
Fix packaging

sni:
Some final changes to get osb builds working.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,6 +1,8 @@
 #!/bin/sh
 
-autoreconf -s -i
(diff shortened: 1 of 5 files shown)


autogen.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-autoreconf -s -i
+autoreconf -s -i || autoreconf -s -i
```
jacobbaungard: I think it would be good to have a comment here mentioning why this is needed.

sni:
you are right, added a comment.

PR #235, opened 2018-06-01
Remove unused parameter

sni:


src/naemon/checks_service.c
```diff
@@ -466,7 +466,7 @@ int handle_async_service_check_result(service *temp_service, check_result *queue
 	else {
 
 		/* parse check output to get: (1) short output, (2) long output, (3) perf data */
-		parse_check_output(queued_check_result->output, &temp_service->plugin_output, &temp_service->long_plugin_output, &temp_service->perf_data, TRUE, FALSE);
+		parse_check_output(queued_check_result->output, &temp_service->plugin_output, &temp_service->long_plugin_output, &temp_service->perf_data, TRUE);
```
roengstrom: Since there might exist NEB-modules that call this function, perhaps we should bump CURRENT_NEB_API_VERSION?

sni: or should we use __attribute__ ((unused)) for now?

roengstrom: I think that sounds like a good idea!

jacobbaungard: Did you check to see if these were the only occurrences of that warning? I suspect that, as we compile with "-Wno-unused-parameter", there could be several places this occurs.

I am fine with using the __attribute__ annotation.

sni:
The only reason for this PR was to support broken gcc versions on sles 11 on obs. Since there are more issues on sles 11, we just drop support for sles 11 obs builds for now.

PR #242, opened 2018-06-05
add obsoletes for naemon-tools

sni:
naemon-tools are now part of the naemon-core project. Adding an obsoletes should
make updates smoother.


rephlex:
sweet!

PR #245, opened 2018-07-09
release 1.0.8

sni:


sni:
i don't think its worth to mention that?

jacobbaungard:
Given it is such a small release log, I thought we might as well, but its OK for me if you think it is not worth to mention.

PR #255, opened 2018-08-28
parse_check: Don't escape already escaped newlines

jacobbaungard:
This patch ensures that newlines which are already escaped in a check results long_output is not escaped again.

When running [Merlin](https://github.com/op5/merlin) to have a distributed Naemon setup, we had a problem where a check result coming from another node, would have double escaped new lines. Merlin sends check results to different nodes, after the check_result has been parsed. The receiving node would in turn also run it through Naemons check result parsing mechanism. This resulted in newlines being escaped twice, once on the remote node, and once on the master (for example). 

This fix works by adding "\\n" to the exception list of g_strescape, which means already escaped newlines won't be escaped again.

This fixes [MON-11112](https://jira.op5.com/browse/MON-11112)


roengstrom:
Nice fix, looks good to me.

Issue #150, opened 2016-09-25
Reload : Fix defunct


Benoit-Forveille:
Hi,

After several reload, there are many defunct process.
in daemon-init, change the reload function

reload() {
    echo -n "Reloading $prog: "
    # pkill -HUP -u ${user} -f ${exec}
    kill -HUP $PID
    retval=$?
    echo
}

hurr1c4n:
Hi,

with my naemon environment I have the same issue.
I solve it by adding the parameter -o.

"-o, --oldest
              Select only the oldest (least recently started) of the matching processes.
"
With this option, I reload the process leader only.

```
reload() {
    log_daemon_msg "Reloading $prog: "
    retval=0
#    pkill -HUP -u ${user} -f ${exec}
    pkill -HUP -u ${user} -f ${exec} -o
    retval=$?
    log_end_msg $retval
}
```

ioguix:
Hi,

We had the same issue here. After moving from Nagios et Naemon (lot of conf to review, lot of reload), we had 363 naemon defuncts processes.

We applied this fix yesterday, restarted Naemon, had a few reload since then, but no more than on defunct at a time according to the monitoring.

dirtyren:
I did this small improvement for my installation.
  In case the pid file does't exist for some reason, it does the old reload and when it reloads, it recreates the PID file, so the new kill -HUP will work on the next reload.

   if [ -e  $pidfile ]; then
        PID=`cat $pidfile`;
        kill -HUP $PID >/dev/null 2>/dev/null
    else
        pkill -HUP -u ${opmon_user} -f ${opmon_bin}
    fi

sciolto:
naemon 1.0.6 on CentOS 6.9, same issue. 

Fixed with @hurr1c4n 's patch 

Thank you !

sni:
does anyone want to put that into a pull request?

dirtyren:
Pull request #216.

jacobbaungard:
Merged (https://github.com/naemon/naemon-core/commit/893d1df40dc4c5649765e26c0edd2f30af7e9b87)

pvdputte:
FYI, I think that this fix is not present in the 1.0.7 and 1.0.8 packages.

jacobbaungard:
I just tested on CentOS 6 (installed from [here](https://labs.consol.de/naemon/release/v1.0.8/rhel6/x86_64/)) and it looks like the patch is included correctly. Which packages did you try?

[...]

PR #256, opened 2018-08-28
Reload : Fix defunct

pvdputte:
Fixes issue #150 for debian/ubuntu packaging as well.

### DIFF
--- a/debian/naemon-core.naemon.init
+++ b/debian/naemon-core.naemon.init
@@ -156,7 +156,12 @@ restart() {
 reload() {
     log_daemon_msg "Reloading $prog: "
     retval=0
-    pkill -HUP -u ${user} -f ${exec}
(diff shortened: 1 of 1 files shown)


debian/naemon-core.naemon.init
```diff
@@ -156,7 +156,12 @@ restart() {
 reload() {
     log_daemon_msg "Reloading $prog: "
     retval=0
-    pkill -HUP -u ${user} -f ${exec}
+    if [ -e $pidfile ]; then
+        PID=cat $pidfile;
```
sni: there should be backticks, right?

jacobbaungard: Yes think so as well, i.e same as: https://github.com/naemon/naemon-core/commit/c523f9c1f2ca4ac0e4097a7c7fa7a5f1c5e23332

PR #257, opened 2018-09-04
Init: Increase delay between SIGTERM and SIGKILL

jacobbaungard:
When using killproc to kill a process, first a SIGTERM signal is sent, and after a default of 3 seconds (5 seconds on SUSE), a SIGKILL is sent.
    
On larger setups we often see that 3 seconds, is not sufficient for Naemon to shutdown, and as a result retention data might not be correctly saved.
    
This commit increase the timeout to 90 seconds on redhat based releases and on SUSE. We only change this one these systems, as killproc seems to differ between different distributions, and might not offer the same arguments. 90 seconds timeout is the default for systemd, and the therefore the rational for choosing that value.

In addition added a log message on save of retention data at shutdown to aid debugging.

More context: [MON-10565](https://jira.op5.com/browse/MON-10565).

### DIFF
--- a/daemon-init.in
+++ b/daemon-init.in
@@ -96,9 +96,11 @@ start() {
 stop() {
     echo -n "Stopping $prog: "
     if [ "x${is_suse}" == "xyes" ]; then
-        killproc -p ${pidfile} -t 10 $exec
+        killproc -p ${pidfile} -t 90 $exec
     elif type start-stop-daemon >/dev/null 2>&1; then
         start-stop-daemon --stop --name ${prog} --pidfile ${pidfile} --exec ${exec}
+    elif [ -e /etc/redhat-release ]; then
+        killproc -p ${pidfile} -d 90 ${exec}
     else
         killproc -p ${pidfile} ${exec}
     fi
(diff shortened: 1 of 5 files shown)


sni:
we should at least wait till there is a solution for #258

jacobbaungard:
For us, this is quite an important issue (i.e retention data not saved correctly), but we can just merge it downstream until #258 is fixed, if you prefer.

Note that this commit, does not impact reload, only stopping/restarting Naemon. The issue described in #258 is not as significant during stopping/restarting due to the fact that the system will kill the process if it hangs during shutdown (after 90 seconds with this commit).

As an alternative we can decrease the timeout to for example 15 seconds instead. 90 Seconds was chosen so that the behavior is the same on init and systemd based systems, but that is not so important perhaps.

roengstrom:
Like @jacobbaungard says, this only applies when restarting/shutting down naemon, so it shouldn't have any affect on #258. And I think that this is an important fix because this means that we get broken retention files on large configurations, which is very bad.

Also, I realize that livestatus is a popular module for naemon, but should fixes in the core be dependent on the behaviour of a module, just as a principle I mean?

Issue #224, opened 2018-05-08
Naemon does not honor the flag use_retained_scheduling_info


jvigna:
When a naemon reload/restart is done if this flag is set naemon should retain the scheduling of all services from before. While this is working in nagios-3.x, in naemon the scheduling is recalculated for all host/service checks. This will lead to a problem if more reloads are done as it may happen that services are not checked for a long time.

jacobbaungard:
Closed by https://github.com/naemon/naemon-core/pull/259

Issue #156, opened 2016-11-01
Host- and Servicechecks maybe never executed


nook24:
Hi,
yesterday i was asking in the IRC channel, if there is a known issue with host rescheduling in Naemon 1.0.3 or 1.0.5.

I noticed that the distance between last_check and next_check in status.dat and retantion.dat are much more than check_interval allows.

For Example:
````
last_check=1477914906 => 12:55:06(CET)
next_check=1477924669 => 15:37:49(CET)
check_invervall=7200
interval_length=1

Diff of the timestams: 9763
````
I also tried with `interval_length=60` but this causes the same issue.
Even i noticed, that `next_check` was calculated new on every Naemon restart/reload but it was not just last_check + check_interval or NOW() + check_interval, the next_check timestamp was set to random values i didn't unterstand.

So i started some investigation and found in `checks_init_hosts ` in `checks_host.c` the following code:
````c
for (temp_host = host_list; temp_host != NULL; temp_host = temp_host->next) {

		/* update status of all hosts (scheduled or not) */
		update_host_status(temp_host, FALSE);

		/* schedule a new host check event */
		schedule_next_host_check(temp_host, ranged_urand(0, get_host_check_interval_s(temp_host)), CHECK_OPTION_NONE);
	}
````
I guess the `ranged_urand(0, get_host_check_interval_s(temp_host))` part is used for load balancing the system, but the random range is way to big, especially in my example where my host has a check interval of 2h.
But the same code is inside of `checks_service.c` and services sometimes have 1d as check_interval, for example for SSL certs.

I only notice this, due to distributed monitoring. At night all systems were running well and report there check results to the naemon master. But on a average working day, the master randomly fires the `check_fresh` command. So i started to take a look on the system and found that next_check > last_check + check_interval. With every naemon restart it get worse and worse 🙈

But i was able to force an example, which is much more worse than the given above:
check_interval=120
````
Naemon Restart:
last_check=1477562797 => Thu Oct 27 12:06:37 CEST 2016
next_check=1477931604 => Mon Oct 31 17:33:24 CET 2016

Naemon Restart
last_check=1477562797 => Thu Oct 27 12:06:37 CEST 2016
next_check=1477936368 => Mon Oct 31 18:52:48 CET 2016
````
As you can see, just with restarting Naemon i can avoid that checks with a long check_interval will be executed, if I am lucky with urand()

nook24:
Here are a few other examples, what happens if i just reload my naemon core a few times:
[...]

PR #259, opened 2018-09-10
Retain next_check schedule on restart (#224, #156)

jacobbaungard:
This commit ensures that the next_check schedule for hosts and services
are retained on Naemon restart, given that use_retained_scheduling_info
is enabled.

The logic is as follows:

- If use_retained_scheduling_info is disabled, set a random time (as
  before)
- If use_retained_schedule_info is enabled:
  - If we didn't miss the check during the restart, retain the old
    next_check time
  - If we missed one check, schedule the service/host within the next
    interval_length (usually 60 seconds)
  - If we missed more than one check, schedule the next check randomly.

We schedule missed checks within 60 seconds, rather than immediately in
order to do some load balacing. This is also the rationale for
scheduling the check randomly, in case we missed more than one check
(this indicates Naemon has been down for a longer period of time).

This fixes:
- https://github.com/naemon/naemon-core/issues/224
- https://github.com/naemon/naemon-core/issues/156
- [MON-10720](https://jira.op5.com/browse/MON-10720)

### DIFF
--- a/src/naemon/checks_host.c
+++ b/src/naemon/checks_host.c
@@ -48,6 +48,8 @@ static int determine_host_reachability(host *hst);
 void checks_init_hosts(void)
 {
 	host *temp_host = NULL;
+	time_t delay;
+	time_t current_time = time(NULL);
 
 	/******** SCHEDULE HOST CHECKS  ********/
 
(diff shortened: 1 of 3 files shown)


sni:
thats great news, thanks

jacobbaungard:
I think this is ready now.

- Added tests
- Fixed an issue from the previous commit (check_interval was in minutes instead of seconds as it should be).

jacobbaungard:
@nook24, @jvigna, how does the logic in this PR sound for you guys? You think it would solve the problems you have reported?

jvigna:
Look very good to me!

jacobbaungard:
I will go ahead and merge this, this afternoon unless there are any objections.

sni:
i was on vacation... :-) no objections.

PR #262, opened 2018-10-01
el6: use correct logrotate script

sni:
Since we install the el7 logrotate in our Makefile.am without further OS
detection we need to replace the logrotate file for el6 later. Otherwise we
would end up with the el7 file and no logrotation.


sni:
thanks for reviewing. Yes, that seems to be the same thing.

PR #263, opened 2018-10-04
Add host_down_disable_service_checks config option

sni:
Add option to *not* check services if their host is down.

references:
      - https://github.com/NagiosEnterprises/nagioscore/commit/05e1dda8ccaa9d633b22b11a555d67724930a3e7

### DIFF
--- a/sample-config/naemon.cfg.in
+++ b/sample-config/naemon.cfg.in
@@ -1024,6 +1024,16 @@ allow_empty_hostgroup_assignment=0
 
 #check_workers=3
 
+
+# DISABLE SERVICE CHECKS WHEN HOST DOWN
+# This option will disable all service checks if the host is not in an UP state
+#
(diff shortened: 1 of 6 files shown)


jacobbaungard:
Looks interesting. Perhaps we should have a unit test for this?

jacobbaungard:
I think this change is good, but would really like to see a short test for it as well. I think we have a reasonable test setup, so we might as well use it (useful for guarding against future regressions etc etc).

sni:
i just didn't have time to look further into this. I will add a simple test case.

jacobbaungard:
This should do, I think. 

[0001-Test-host_down_disable_service_checks.txt](https://github.com/naemon/naemon-core/files/2663322/0001-Test-host_down_disable_service_checks.txt)

sni:
thanks, i will add that one

PR #272, opened 2018-12-03
fix memory leak when overriding checks

sni:
Overriding checks during the host/service_initiate stage leads to a memory leak.
Freeing the check_result pointer helps.

### DIFF
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -307,6 +307,7 @@ static int run_scheduled_service_check(service *svc, int check_options, double l
 	if (neb_result == NEBERROR_CALLBACKOVERRIDE) {
 		clear_volatile_macros_r(&mac);
 		free_check_result(cr);
+		nm_free(cr);
 		nm_free(processed_command);
 		return OK;
 	}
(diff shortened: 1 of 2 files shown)


src/naemon/checks_service.c
```diff
@@ -307,6 +307,7 @@ static int run_scheduled_service_check(service *svc, int check_options, double l
 	if (neb_result == NEBERROR_CALLBACKOVERRIDE) {
 		clear_volatile_macros_r(&mac);
 		free_check_result(cr);
+		free(cr);
```
jacobbaungard: shouldn't `free_check_result` handle that?

sni: would be ok for me, i just just checked how the other parts handle freeing check_results.

jacobbaungard: Okay, makes sense, thanks for taking a look.

Perhaps we should use nm_free though?

sni: you are right, i pushed a new single commit fixing all this

jacobbaungard: Great, looks fine for me once the build passes.

PR #274, opened 2018-12-10
Add host down disable service checks

sni:
seems like the original upstream PR #263 branch has been removed meanwhile, here is a new one including the test case. thanks @jacobbaungard


sni:
thanks

Issue #287, opened 2019-03-29
Naemon fake state change events if the host is down


ibering:
Today my colleagues discovered a weird behavior with
"fake" a hard state change for the services.

### What's the issue with this?
This manipulates (and breaks) the data of `nebstruct_statechange_data`.

| Time | Check # | State    | StateType | State Change | Notes                                                                                                                    |
|------|---------|----------|-----------|--------------|--------------------------------------------------------------------------------------------------------------------------|
| 0    | 1/5     | DOWN     | SOFT      | YES          | Host change from HARD Up to Down SOFT - OK                                                                               |
| 1    | 1/3     | Critical | SOFT      | YES          | Service change from Ok HARD to Critical SOFT - OK                                                                        |
| 2    | 2/3     | Unknown  | HARD      | YES          | Services change from Critical SOFT to Unknown HARD. `last_hard_state` gets set to 3 - ERROR should be SOFT state and last_hard_state still be 0|

![Bildschirmfoto 2019-03-29 um 12 04 41](https://user-images.githubusercontent.com/9592910/55228889-a6f24280-521b-11e9-8f86-7c96654fbba2.png)

### Expected behavior
| Time | Check # | State    | StateType | State Change | Notes                                                                                                            |
|------|---------|----------|-----------|--------------|------------------------------------------------------------------------------------------------------------------|
| 0    | 1       | DOWN     | SOFT      | YES          | Host change from HARD Up to Down SOFT                                                                            |
| 1    | 1/3     | Critical | SOFT      | YES          | Service change from Ok HARD to Critical SOFT                                                                     |
| 2    | 2/3     | Unknown  | SOFT      | YES          | Service change from Critical SOFT to Unknown SOFT -Set SOFT State, dont touch last_hard_state                    |

This only happens if the host is in a non Up state.

If the host is Up, the last_hard_state works as expected:

![Bildschirmfoto 2019-03-29 um 12 04 47](https://user-images.githubusercontent.com/9592910/55228910-bd989980-521b-11e9-9943-0eae7c87e929.png)

I guess this is the code which is causing the issue:
https://github.com/naemon/naemon-core/blob/1b66a24c4b7753e5d0bafded434cb1c0ca026457/src/naemon/checks_service.c#L834-L857

This looks wrong to me: `temp_service->last_hard_state = temp_service->current_state;`

jacobbaungard:
This is by design basically. The service status is changed to HARD if the associated host is in (iirc) hard down. 

[...]

PR #291, opened 2019-04-10
Issue 287

nook24:
This should resolve issue #287 without breaking the Naemon logic.

Current behavior:
default naemon no patch

With this patch:
patched_naemon

Also some more state changes:
rainbow_test


jacobbaungard:
If I understand it correctly this patch changes when `last_hard_state` is set, so that it is set after the NEB event handler callback that I assume you use to capture the results?

I see a couple of problems with this. 

1. The meaning of `last_hard_state` is now inconsistent. That is for a period of time `last_hard_state` != current_state when `state_type` is hard

2. The meaning of `last_hard_state` has now changed for other modules which are using the event handler callback

I would much prefer implementing this behavior as a separate parameter - although obviously that is a lot more work.

ibering:
Isn't it now the same behavior like if the host is up?

https://github.com/naemon/naemon-core/blob/72aa5cbff91e994fd61383ea55ded407bd48f2f1/src/naemon/checks_service.c#L988-L993

Host is Up:
Call `handle_service_event()` and change `last_hard_state`

Host is down (without patch)
Change `last_hard_state`
https://github.com/naemon/naemon-core/blob/72aa5cbff91e994fd61383ea55ded407bd48f2f1/src/naemon/checks_service.c#L836-L857
 and call `handle_service_event()`
https://github.com/naemon/naemon-core/blob/72aa5cbff91e994fd61383ea55ded407bd48f2f1/src/naemon/checks_service.c#L881-L893
so `handle_service_event()` gets called to late I think.

---

With the patch the behavior of Naemon is the same doesn't matters if the host is up or not up. In the current implementation, there are to different behaviors.

sni:
without having looked into details, i agree with @jacobbaungard. Also setting `last_hard_state_change` but not the actual state definitly results in an inconsistant wrong state where `last_hard_state_change` points to the current state, but the state itself is the old one.

ibering:
But it is already inconsistent now. Why is there different behavior between host UP and host NOT UP? Please see screenshots with "Current behavior" :/

sni:
don't get me wrong, all i am saying is, that if you move setting `last_hard_state`, move `last_hard_state_change` along with it.

nook24:
> don't get me wrong, all i am saying is, that if you move setting last_hard_state, move last_hard_state_change along with it.

Yes that makes sense. I've fixed this

jacobbaungard:
I missed that there is a different logic when the host is OK - that is not good.

I guess we need to decide what `last_hard_state` should mean. I.e. should it be the _previous_ hard state or the _latest_ hard state (that is including the `current_state` if `state_type` is hard), and then implement a behavior which is consistent across all states.

I always thought it as as the _latest_ hard state.

With the latest commit it is better, but I find it strange to pass some values to the event handler for them then to be immediately changed afterwards.

jacobbaungard:
But now, when there is a hard state change to OK, we don't actually adjust `last_hard_state` at all. So that should probably be fixed as well?

sni:
if thats the case, this should be fixed of course :-)

nook24:
In our case we use the data provided by `broker_statechange_data` to build up a [state history](https://github.com/statusengine/worker/blob/master/lib/mysql.sql#L33-L52) for all hosts and services.

If a host is up `last_hard_state` always holds the `previous` hard state when `broker_statechange_data` gets called by Naemon.

So for me, I thought last_hard_state == previous_hard_state. Which also caused some misunderstandings in the conversation above and #291 :)

If a host is down, `last_hard_state` behaves slidely different when `broker_statechange_data` gets called it holds the current hard state. So for me this was a bug. It also makes sense for me that if the system give me a statechange event, that this event contains the previous state.

I know `last_hard_state` is also used in service status events. I never checked if it holds there the "current hard state" or the "previous hard state"...


Enjoy the weekend :)

jacobbaungard:
While I think it is a bit odd to do the change after the after the event handler is called, I am ok with this as long as the behavior for the UP state is the same. So perhaps you could add the same logic for when there is a hard state change to UP (around L732 I believe it should be) ?

nook24:
@jacobbaungard 
Unfortunately I'm not sure what I should do^^

Case "Services goes from HARD Critical to HARD Ok (Host is UP)":
- Call to `handle_service_event()` in [L732](https://github.com/naemon/naemon-core/blob/143f7ef6aaff5a59261c1499371a38dfd0e12089/src/naemon/checks_service.c#L732)
- Set `last_hard_state` to Ok in [L764](https://github.com/naemon/naemon-core/blob/143f7ef6aaff5a59261c1499371a38dfd0e12089/src/naemon/checks_service.c#L764)

Case "Service goes from HARD Ok to HARD Critical (Host is UP)"
- Call to `handle_service_event()` in [L910](https://github.com/naemon/naemon-core/blob/143f7ef6aaff5a59261c1499371a38dfd0e12089/src/naemon/checks_service.c#L910)
- Set `last_hard_state` in [L995](https://github.com/naemon/naemon-core/blob/143f7ef6aaff5a59261c1499371a38dfd0e12089/src/naemon/checks_service.c#L995)

Or should I move [L650](https://github.com/naemon/naemon-core/blob/143f7ef6aaff5a59261c1499371a38dfd0e12089/src/naemon/checks_service.c#L650) around?

jacobbaungard:
Merged this, thanks!

PR #294, opened 2019-05-08
fix expanding service dependencies

sni:
when iterating over parents and childs during service dependency expansion, we
need to reset the children map for each parent. Otherwise dependencies are
only added for the first parent while it should be added for all parents.


sni:
i added a test case and also move the bitmap_clear a bit further.

PR #295, opened 2019-07-03
Fix format overflow

immae:
The compiler is complaining about some pointers or strings that may be
null.
NB: some pointers are actually null (for instance temp_ptr which has
some == NULL checks), so the message could probably be rephrased.


sni:
looks good, thanks

PR #296, opened 2019-07-09
add last_update timestamp to track host/service changes

sni:
This PR adds a new attribute last_update to hosts and services. This timestamp
will be updated before and after servicechecks, on notifications and on
external commands. Once this timestamp is accessible via livestatus, it will
make synchronisation (ex. by LMD) easier.

### DIFF
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -262,6 +262,11 @@ static int run_scheduled_service_check(service *svc, int check_options, double l
 		               svc->description, svc->host_name, svc->id,
 		               neb_result == NEBERROR_CALLBACKCANCEL ? "cancelled" : "overridden");
 	}
+
+	/* get the command start time */
+	gettimeofday(&start_time, NULL);
+	svc->last_update = start_time.tv_sec;
+
 	/* neb module wants to cancel the service check - the check will be rescheduled for a later time by the scheduling logic */
 	if (neb_result == NEBERROR_CALLBACKCANCEL) {
 		return ERROR;
@@ -296,9 +301,6 @@ static int run_scheduled_service_check(service *svc, int check_options, double l
 		log_debug_info(DEBUGL_CHECKS, 0, "Processed check command for service '%s' on host '%s' was NULL - aborting.\n", svc->description, svc->host_name);
 		return ERROR;
 	}
-
-	/* get the command start time */
-	gettimeofday(&start_time, NULL);
 
 	cr = nm_calloc(1, sizeof(*cr));
 	init_check_result(cr);
@@ -1051,6 +1053,7 @@ int handle_async_service_check_result(service *temp_service, check_result *queue
 	nm_free(old_plugin_output);
 	nm_free(old_long_plugin_output);
 
+	temp_service->last_update = current_time;
 	return OK;
 }
 
--- a/src/naemon/checks_host.c
+++ b/src/naemon/checks_host.c
@@ -253,6 +253,7 @@ static int run_async_host_check(host *hst, int check_options, double latency)
 	start_time.tv_usec = 0L;
 	end_time.tv_sec = 0L;
 	end_time.tv_usec = 0L;
+	hst->last_update = now;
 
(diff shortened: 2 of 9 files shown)


src/naemon/checks_service.c
```diff
@@ -242,6 +242,9 @@ static int run_scheduled_service_check(service *svc, int check_options, double l
 	int macro_options = STRIP_ILLEGAL_MACRO_CHARS | ESCAPE_MACRO_CHARS;
 	int neb_result = OK;
 
+	/* get the command start time */
+	gettimeofday(&start_time, NULL);
```
jacobbaungard: Do we need this? It is reset to `0L` a few lines down.

As a result of moving this here, before the initialization below, do we not always set that check result start & endtimes to 0 effectively ? (L313)

sni:
thanks for pointing this out. `last_update` should only be reset to zero on restarts. That was by mistake. I moved the gettimeofday again. Should be fine again now. 
Let me double check this, but my observation is that last_update and last_check should not have the same value for any check that takes more than 1 seconds. It is set to the start time of the check in
https://github.com/naemon/naemon-core/blob/master/src/naemon/checks_service.c#L460
The reason to update last_update when the check starts is that i want to catch the update of the `is_executing` flag.

jacobbaungard:
Why do you want to reset `last_update` to 0 on restarts? The value is still valid after restarts if general state retention over restarts is enabled?

sni:
tbh, i don't care if it is reset to zero or not since LMD will refetch everything after a reload anyway. But if there is a use case for keeping the value, i can add this as well.

jacobbaungard:
I think we should keep it over restarts if state retention is enabled. I don't think we have any other fields that are not retained, if the retain settings are enabled.

sni:
ok, i added retention handling.

jacobbaungard:
Thanks.

Looks better. But I still get that last_update and last_check are the same after execution of the check. I.e:
```
# echo "GET services\nColumns: host_name description last_check last_update execution_time\nFilter: description = slow" | unixcat /opt/monitor/var/rw/live_tmp
monitor;slow;1562765853;1562765853;5.0039770000e+00
# printf "[%lu] SCHEDULE_SVC_CHECK;monitor;slow;%lu\n" `date +%s` > /opt/monitor/var/rw/nagios.cmd
# echo "GET services\nColumns: host_name description last_check last_update execution_time\nFilter: description = slow" | unixcat /opt/monitor/var/rw/live_tmp
monitor;slow;1562765853;1562765966;5.0039770000e+00
# echo "GET services\nColumns: host_name description last_check last_update execution_time\nFilter: description = slow" | unixcat /opt/monitor/var/rw/live_tmp
monitor;slow;1562765971;1562765971;5.0040860000e+00
```
Slow is running the following check plugin:

```
#!/bin/bash
sleep 5
echo "OK"
exit 0
```

sni:
strange, i cannot reproduce that. last_update and last_check are different here. Do you have any modules loaded (besides livestatus)?

Doing this query while a `sleep 3` check is running:
`GET services\nColumns: state has_been_checked is_executing last_check last_update\nFilter: description = slow\n"`

results in:
```
0;0;0;0;0
0;0;1;0;1562834995
0;0;1;0;1562834995
0;0;1;0;1562834995
0;1;0;1562834995;1562834998
0;1;0;1562834995;1562834998
```

So the last_check is set to the start date of the check immediatly when the check starts. When finished, last_update is set to the end time of the check. Thats exactly what i would expect.

jacobbaungard:
Seems like merlin is disrupting this somehow. Investigating.

jacobbaungard:
Interesting. Merlin actually changes the `last_check` so that it update when the check is being processed: https://github.com/ITRS-Group/monitor-merlin/blob/d7ba3a5ac2f6e92db46dc2917b8403b8b092ddf3/module/hooks.c#L356

As a result the last_check and last_update is the same post check.

sni:
might be interesting to see that behaviour in vanilla naemon as well.

PR #301, opened 2019-08-19
bail out if members cannot be expaned (fixes #300)

### LINKED ISSUE #300: Naemon not correctly handling configuration error.
Hello,

When
 
   * a host group contains hosts that do not exist,
   * an error detected, can see the error message in the configuration check output.  See line 'Error: Could not find any host matching 'isilon-dev-f-1.its' (config file '/etc/socg/users/danny/conf.d/services/auto-checks/des-systems/soe/allservers.cfg', starting on line 1)'
   * no error is reported in summery, see 'Total Errors:   0'
   * exit val from naemon is 0.

OS: RHEL7.3
Naemon Version: 1.0.6
RPM: 1.0.6.1.el7 from consol labs.
Attached: naemon.txt.  script output from running naemon in validation mode.

[naemon.txt](https://github.com/naemon/naemon/files/1079404/naemon.txt)

sni:
Configuration validation should bail out if host members cannot be found. Right now it would
print an error but continue. Problem here is that the rest of the list is not parsed, so the
first unknown host cuts off the members list.
Same applies to servicegroup members.

### DIFF
--- a/src/naemon/xodtemplate.c
+++ b/src/naemon/xodtemplate.c
@@ -5029,6 +5029,7 @@ static int xodtemplate_recombobulate_hostgroups(void)
 	xodtemplate_hostgroup *temp_hostgroup = NULL;
 	char *hostgroup_names = NULL;
 	char *ptr, *next_ptr, *temp_ptr = NULL;
+	int res;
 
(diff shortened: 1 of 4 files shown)


jacobbaungard:
Looks good, but should be rebased (hand't noticed you had included all the testcase fixes to this PR, too)

PR #312, opened 2019-08-20
fix downtime tests

sni:
it is unsafe to access the downtimes after they have been removed because this
frees the underlaying memory. This might work sometimes if the memory has not
been reused meanwhile but is undefined behaviour.
The effect of has been asserted indirectly through the affected hosts/service already.


tests/test-scheduled-downtimes.c
```diff
@@ -266,7 +262,6 @@ START_TEST(host_multiple_fixed_scheduled_downtimes_one_cancelled_one_stopped)
```
jacobbaungard: Do we need the one just above this line as well? Given that we call `unschedule_downtime` before accessing `is_in_effect`.

sni: you are right, missed that one.

Issue #162, opened 2017-01-17
NEBCALLBACK_SERVICE_STATUS_DATA and NEBCALLBACK_HOST_STATUS_DATA are send twice to broker modules


nook24:
Hi,
i have a quick question an maybe i just had implemented this wrong.
My broker module is consuming `NEBCALLBACK_SERVICE_STATUS_DATA` and `NEBCALLBACK_HOST_STATUS_DATA` events from Naemon Core.

Here is the data i get, after every service check: https://gist.github.com/nook24/bb913d5f94857690ece142bf289fb9f1
As you can see, the first and the second call are the same data.
````C
servicestatusdata = (nebstruct_service_status_data *)data
servicestatusdata->type
````
Even the type is always `1202` and the timestamps are the same.

The problem i have since a long time is, that Naemon call this callback twice per service check. I found a bug in http://tracker.nagios.org/view.php?id=4 from 2009 with the comment `...Modules can trap this and handle what they want accordingly.`

So my question, is there an other event i should use? 
To filter this duplicates events is a huge job in C so i need to crate a hash map, crate a hash of every host and service status struct and need to check if the hash sum is inside my hash map, or some stuff like this.

Many thanks :)

tvestelind:
Hi,

Is your code available somewhere so that I could look at it?

I'm not sure what would be the problem but maybe I could troubleshoot a bit :)

nook24:
Hi @tvestelind,
yes, i added you a link to the source.
This is where the callbacks are registered:
https://github.com/nook24/statusengine/blob/master/statusengine/src/statusengine.c#L286-L287

Every naemon event gets encoded as json objects and send to a queue:
https://github.com/nook24/statusengine/blob/master/statusengine/src/statusengine.c#L679-L760

So the host and service status records are the only callbacks with this duplicate data behavior...

tvestelind:
Hi,

Sorry, I haven't had time to test this. I'm keeping it in mind but I'm quite busy a.t.m. :/

sni:
`update_service_status` gets called twice. One time when receiving the check result and another time when it gets rescheduled. The question here is what do you want to achieve and why is this an issue?

nook24:
@sni, sorry I totally missed your reply...
I only want to save the check result.

PR #313, opened 2019-09-02
release 1.1.0

sni:


nook24:
```
* add internal last_update timestamp to track host/service changes
```
Could this be used to resolve #162?

sni:
i don't think so. The last_update timestamp is useful when you want to fetch data, the issue above is when you want to push data.

PR #317, opened 2019-10-02
fix E: invalid-license

sni:
according to https://spdx.org/licenses/ the official license tag is "GPL-2.0-or-later". Otherwise
rpmlint complains about "E: invalid-license"

### DIFF
--- a/naemon-core.spec
+++ b/naemon-core.spec
@@ -11,7 +11,7 @@ Summary: Open Source Host, Service And Network Monitoring Program
 Name: naemon-core
 Version: 1.1.0
 Release: 0
-License: GPLv2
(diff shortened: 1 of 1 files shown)


jacobbaungard:
Shouldn't it be `GPL-2.0-only` then? Or do we allow GPLv3 ? (I don't have any strong preferences).

sni:
i have no strong feelings either, i always understood gplv2 as gplv2 or later but i really don't care :-)

jacobbaungard:
I think that everyone would need to explicitly agree to "later", so might be safer to keep it at v2.

sni:
ok, changed it.

PR #322, opened 2020-01-21
Retain flap detection option over restarts

jacobbaungard:
This commit ensures the flap detection of a host/service is retained
over restarts, when changing the flap detection option using an external
command. This is in line with other options such as passive checks,
notifications etc.


sni:
thanks

PR #323, opened 2020-02-17
Feature skip check options

sni:


src/naemon/utils.c
```diff
@@ -166,6 +166,10 @@ char *use_timezone = NULL;
 int allow_empty_hostgroup_assignment = DEFAULT_ALLOW_EMPTY_HOSTGROUP_ASSIGNMENT;
 int allow_circular_dependencies = DEFAULT_ALLOW_CIRCULAR_DEPENDENCIES;
 int host_down_disable_service_checks = DEFAULT_HOST_DOWN_DISABLE_SERVICE_CHECKS;
+int service_skip_check_dependency_status = -1;
```
jacobbaungard: Could we use defines here, in line with the other settings?

PR #329, opened 2020-03-22
fix crash in DEL_DOWNTIME_BY_HOSTGROUP_NAME

kjetilho:
uninitialised struct causes dereference to random addresses.

I first thought this was due to supplying too few parameters (Icinga 1 only requires one, the hostgroup), but this command would never work, I think.

### DIFF
--- a/src/naemon/commands.c
+++ b/src/naemon/commands.c
@@ -1776,13 +1776,13 @@ static void foreach_contact_in_contactgroup(contactgroup *target_contactgroup, v
 }
 
 struct matches_arg {
-	struct external_command *ext_command;
+	const struct external_command *ext_command;
 	int deleted;
 };
 static gboolean delete_if_matches(gpointer _name, gpointer _hst, gpointer user_data)
 {
 	struct matches_arg *match = (struct matches_arg *)user_data;
-	struct external_command *ext_command = match->ext_command;
+	const struct external_command *ext_command = match->ext_command;
 	host *host_ptr = (host *)_hst;
 	if (strcmp(GV_STRING("hostname"), "") && !strcmp(host_ptr->name, GV("hostname")))
 		return FALSE;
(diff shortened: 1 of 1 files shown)


jacobbaungard:
Thanks a lot for sending this in. I tested this and it definitely fixes the crash which is great, but I still seem to be unable to remove any downtimes with this command.

Did you manage to get it to work as is?

I think we need some testing for these functions to make sure they work correctly.

kjetilho:
you are absolutely right!  haha.

unfortunately, this is sort of on purpose.  in src/naemon/commands.c:1101

```c
	/* Do not allow deletion of everything - must have at least 1 filter on */
	if (hostname == NULL && service_description == NULL && start_time == 0 && cmnt == NULL)
		return deleted;
```

looks like the function needs an override for this case?

kjetilho:
I think I found a better fix (it is better when it is local).  it now works to delete all downtimes for a hostgroup.  It is also backwards compatible with Icinga 1 since the missing parameters are filled with defaults which the filter understand as unset.

jacobbaungard:
When I tried, I didn't manage to delete any downtimes with `DEL_DOWNTIME_BY_HOSTGROUP_NAME` even if I supplied all the variables.

I am not sure the dummy hostname will work correctly. When supplying a dummy name host to `delete_downtime_by_hostname_service_description_start_time_comment` I would assume it will fail to find any downtimes, because there is no host with the same name as the dummy name ?

kjetilho:
the hostname is not really a dummy, it is the hostname for each member host in the hostgroup, so it is guaranteed to match.

it is still possible to supply a hostname filter in the livestatus command, and this will be respected.  this does not make much sense, since you could use DEL_DOWNTIME_BY_HOST_NAME instead.  at best it is a failsafe, since both hostgroup and hostname must match for the command to have effect.

PR #333, opened 2020-07-13
update debian compat level

sni:
Compatibility levels before 10 are deprecated with current debian testing.


sni:
in case you are wondering, i removed the if clause from the version.sh, since debian runs a autoreconfigure now and would tag all builds as source build anyway.

PR #337, opened 2020-08-11
Worker: Null pointer sanity check before wlog call

jacobbaungard:
This commit implements a couple of sanity checks on pointers in the
worker process. This is the result of analyzing core-dumps from a crash.


sni:
Would it makes sense to still print/log the error, even if cp is not set, but without those details?

jacobbaungard:
Yes makes sense. Amended.

sni:
thanks, merged

PR #338, opened 2020-08-14
correct acknowledgement external command help

sni:
the current documentation is plain wrong here. The related comment for acknowledgements always survive restarts of the core.
The peristent flag just sets if the comment survices removal of the acknowledgement. The flag is checked in
delete_host_acknowledgement_comments() and delete_service_acknowledgement_comments()


sni:
Once this is merged, we need to run the update script from the naemon.org page as well.

PR #341, opened 2020-10-08
Fix indentation

akselsjogren:
This PR fixes paths to sources in the `indent-all.sh` script and reindent the entire tree.

Since paths wasn't correct in `indent-all.sh`, indentation has been lacking behind the rules since sourced were moved around in 7344116806274b. Therefore this is a large change, re-indenting everything according to the project rules.


sni:
Looks good. Can we add a test case on travis to check if new commits are formated correctly?

akselsjogren:
That sounds like a good idea.
I made changed to `indent-all.sh` to support that. Unfortunately the astyle version available on CI (ubuntu xenial) is old (2.x) and complains on a few things that astyle 3.1 has formatted nicely on my machine. It looks like bugs in astyle, cause it doesn't make sense.
I tried to change travis to ubuntu bionic instead, which has the latest astyle (3.1). But then a timeperiod testcase failed :(
It it this test: https://github.com/naemon/naemon-core/blob/70944db34113e41d9ebe621c75c320ba1956d930/t-tap/test_timeperiods.c#L380
From `test-suite.log`:
```
ok 93 - Testing exclude always                                                                                                                                                                                                                                 
ok 94 - 12 Jul 2010 15:00:00 should not be valid                                                                                                                                                                                                               
not ok 95 - There should be no next valid time, was Mon Nov  1 00:00:00 2038                                                                                                                                                                                   
                                                                                                                                                                                                                                                               
#     Failed test (t-tap/test_timeperiods.c:main() at line 380)                                                                                                                                                                                                
ok 96 - Testing Sunday 00:00-01:15,03:15-22:00
```
I tried to understand why it happens, but couldn't. It's a newer gcc on bionic which might have something to do with it? I tried with gcc 7.3 on CentOS 7, but there it works.

sni:
Let me have a look on a bionic test machine. Maybe its some 32bit thing.

akselsjogren:
* Removed running the indent check script in CI for now.
* Rebased branch.

All changes in the big commit are made by *astyle* when running `./indent-all.sh`.

PR #345, opened 2020-11-11
fix next_check synchronisation

sni:
last_update wasn't updated in some cases when only the next_check advances. For example for
pending and disabled hosts/services.
This patch ensures, that whenever next_check is updated, also last_update is set.

### DIFF
--- a/src/naemon/checks_host.c
+++ b/src/naemon/checks_host.c
@@ -122,6 +122,7 @@ void schedule_next_host_check(host *hst, time_t delay, int options)
 	/* Schedule the event */
 	hst->check_options = options;
 	hst->next_check = delay + current_time;
+	hst->last_update = current_time;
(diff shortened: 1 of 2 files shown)


jacobbaungard (approved):
Looks likes this will set `last_update` to the current time when for example scheduling a check via an external cmd? I.e `SCHEDULE_HOST_CHECK` . 

So the question is, do we think that changing the `next_check` is an counts as an "update" ? Seems sensible to me.

There are no definitions of `last_update` anywhere. Perhaps we should add it to the livestatus docs.

sni:
There is a short sentence about the attribute in livestatus:
`Time of the last update of this host (Unix timestamp)`
The `last_update` should basically be updated whenever any attribute of the host/service changes. We set it for external commands already, we just missed some cornercases.

sni:
Here is a PR to update the tables descriptions to match the ones on naemon.org https://github.com/naemon/naemon-livestatus/pull/82

PR #347, opened 2020-11-27
debian: move to systemd

sni:
The oldest supported system is ubuntu14.04 on obs right now and even that supports systemd already. So
simply move to systemd completely. Detection of system failed on obs on some systems and the packages
end up with half complete systemd integration.

dh_installinit install service anyway, even if we do not add the other systemd files.

  [  144s] dh_installinit --name=naemon
  [  144s] 	install -d debian/naemon-core/lib/systemd/system
  [  144s] 	install -p -m0644 debian/naemon-core.naemon.service debian/naemon-core/lib/systemd/system/naemon.service

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,8 @@ Source: naemon-core
 Section: net
 Priority: optional
 Maintainer: Naemon Core Development Team <naemon-dev@monitoring-lists.org>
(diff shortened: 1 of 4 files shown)


jacobbaungard (approved):
Looks good!

PR #350, opened 2021-01-07
Fix header file location in deb/rpm packages

### LINKED ISSUE #299: /usr/include/naemon/naemon/ instead of /usr/include/naemon
Hello,

I was moving some apps from CentOS6 to  CentOS7 and I noticed that naemon-devel-1.0.6 installs its headers under /usr/include/naemon but the latest package 1.0.10 naemon-devel installs under /usr/include/naemon/naemon/ .

   If there any reason for this change?

  Tks.

atj:
Automake automatically adds a package name suffix to `includedir` so it
should be specified as "/usr/include" not "/usr/include/naemon".

Fixes #299.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -12,7 +12,7 @@ override_dh_auto_configure:
 				--bindir="/usr/bin" \
 				--datadir="/usr/share/naemon" \
 				--libdir="/usr/lib/naemon" \
-				--includedir="/usr/include/naemon" \
+				--includedir="/usr/include" \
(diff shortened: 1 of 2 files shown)


sni:
great, thanks.

atj:
Thanks. Could you trigger a rebuild of the packages on OBS?

sni:
https://build.opensuse.org/project/show/home:naemon:daily are triggered automatically, but it might take a few moments. Stable builds are triggered manually upon release.

PR #351, opened 2021-01-08
dh-systemd has been merged into debhelper

sni:
it is obsolete since stretch and removed in debian 11.


jacobbaungard (approved):
Looks good to me

PR #353, opened 2021-02-09
Fix stopping Naemon in logging feature test

akselsjogren:
We noticed that logging tests failed when building in a new environment, it seemed to be a timing issue where the test case wants to SIGTERM naemon, but some investigation gave that naemon didn't completely start yet, so its signal handlers weren't installed and thus didn't stop as intended. This is an attempt to improve the timing issue.

* Fix test that failed if Naemon haven't yet started correctly and got
  it's signal handlers setup at the stage a test case executes the "When
  I stop naemon" step.
  Add a step that starts Naemon and wait until it finds a line in the
  log file that indicates that it should be ready.
  It's not perfect, but it seems to make the test case more stable.

* Close opened files in feature tests.

Signed-off-by: Aksel Sjögren <asjogren@itrsgroup.com>

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -1,5 +1,6 @@
 from support.naemon_object_config import NaemonObjectConfig
 from support.naemon_system_config import NaemonSystemConfig
+from support import slurp_file
 import tempfile
 import os
 import os.path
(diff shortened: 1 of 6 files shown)


sni:
Is that the reason why the travis tests fail occasionally? For example this one: https://travis-ci.com/github/naemon/naemon-core/builds/213749004

akselsjogren:
> Is this the reason why the travis tests fail occasionally? For example this one: https://travis-ci.com/github/naemon/naemon-core/builds/213749004

That is the exact same failure I had.

sni (approved):
looks good

PR #354, opened 2021-02-23
speed up cleanup() when having servicegroups

sni:
while clean up we iterate over all services and each service is removed from
all its servicegroups by iterating over the servicegroup members over and over
to remove each service one by one.
Adding a new flag truncate_lists which can be set on teardown to simply iterate once
over each list and remove/free all items.
This reduces the duration of a config check for an example config with 100k services from
30seconds to less than 3seconds.


jacobbaungard (approved):
Nice! :+1:

PR #355, opened 2021-02-24
Correclty set svc check_command for broker call

jacobbaungard:
This commit fixes a bug where the `check_command` argument was not
correctly set for the calls to `broker_service_check` with the
`NEBTYPE_SERVICECHECK_PROCESSED` type.

This in turn meant that the `command_` members of the
`nebstruct_service_check_data` struct was never populated.

This is similar to the behaviour for host checks.

Signed-off-by: Jacob Hansen <jhansen@itrsgroup.com>
Co-authored-by: Robert Wikman <rwikman@itrsgroup.com>


sni (approved):
looks good

Issue #362, opened 2021-05-25
Unique identifier for notifications


nook24:
Hi,

I'm looking for an method to have an unique identifier for a notification that got sent to multiple contacts.

Unfortunately the macros `$HOSTNOTIFICATIONID$` and `$SERVICENOTIFICATIONID$` depend on the `retention.dat` and are therefore not really unique.

My next idea was to concat `$HOSTNAME$ + $SERVICEDESC$ + $TIMET$` to get a unique id. I checked the Naemon source code and for me it looks like that the value of `$TIMET$` get calculated when the command line of the notification script gets assembled. So the value of `$TIMET$` could be 12:00:00 for one contact and 12:00:01 for the second contact.

My suggestion would be, to add a new UUID macro to the notification logic. Same as:
https://github.com/naemon/naemon-core/blob/6d3ec3405d1c14129f59eb42e7f059ada7045351/src/naemon/notifications.c#L1362

Like so:
```C
nm_asprintf(&mac.x[MACRO_HOSTNOTIFICATIONUUID], "%s", get_uuid_v4()); 
```
This would ensure that all notifications would have an unique identifier. Before I start implementing this and send over a pull request I wanted to ask if this is a change you would merge into the core or not. I would probably make use of `uuid/uuid.h`. So this would add in a new dependency.

sni:
tbh, i am not a big fan of adding a new dependency to solely solve a single very specific detail. Naemon uses a notification id internally already which is unique, even over restarts. As you noticed already, state is stored in the retention.dat file. If you really
need a global unique id, you could create a uuid and create a lookup map from the internal notification id to your uuid.
I assume you want to use it in your statusengine neb module?

nook24:
I admit that my request is an edge case. For me it would be good if I could use some `INSERT ... ON DUPLICATE KEY UPDATE` syntax inside of my notification scripts. The issue with the current notification_id is, that in case of a rollback of the System (ZFS replication, or VM snapshots for example) it could happen that I end up with an wrong / old notification_id which was already used before that rollback happened. But tbh it's more likely that Naemon crashes and the `retention.dat` is old and I end up with duplicate notification_id's.

I understand why you try to keep Naemon as free of dependencies as possible. An dependency free version which would be still unique enough would be something like this:
```C
#include <time.h>
#include <stdlib.h>
#include <stdio.h>

int main() {

    int i;
    int notification_id = 1;    // svc->current_notification_id
    
    srand(time(NULL));

[...]

PR #366, opened 2021-06-12
feature vault macros and broker

sni:
this PR implements dynamic macro expansion from neb modules in order to
implement password vault broker neb modules. From naemons point of view, we
simply define a new macro prefix $VAULT...$ and add broker callbacks to fill
the value.

In order to make password vault work, we need enter a master password before
naemon starts, thats why closing stdin moved to after the neb module
initialization. So the neb module may implement reading something from stdin.

Advantage of this vault macros would be:

  - dynamic 3rd party macro expansion is not limited to passwords
  - password storage can be implemented in any way you like, ex. simply encrypted file or advanced remote vaults
  - vault macros are not limited to numbers like $VAULT1$ but can be anything ex.: $VAULTSNMPCOMMUNITY$
  - some context might be interesting during macro expansion, something like hostname, etc...
    thats why the macros *mac struct is passed to the broker which might contain host/service/contact pointer.

Things to be done:

  - [x] master password is lost during reloads and stdin is closed at that point,
    so need to find a way to save the master password over reloads.
  - [x] systemd integration needs to be tested (can stdin be used here)
  - [x] add simple example neb module


sni:
there is a POC module written in go here: https://github.com/sni/naemon-vault-example/blob/master/main.go

jacobbaungard:
Nice!

Some thoughts:
- Maybe it should be named something else than vault, given it can be used for anything (not just passwords)
- How to handle naming collisions if multiple NEB modules implements this (or limit to only have one NEB module allowed with that capability)
- When we do a reload we don't (have to) clear all memory, so we could probably implement some method for NEB modules to initialize memory on first load which isn't reset. Will probably be a bit messy and that wouldn't work on full restarts of course.
- Another solution could be to store the vault secret in a root-only readable file on disk, I believe one could then make use of `StandardInput=file:/path/to/credentials` in systemd but I have not tested this. Slightly less secure than using user input directly, but saves one from having to do things manually.

sni:
Thanks for having a look.

>     * Maybe it should be named something else than vault, given it can be used for anything (not just passwords)
Any name will do. I chose VAULT because that was the main purpose for now.

>     * How to handle naming collisions if multiple NEB modules implements this (or limit to only have one NEB module allowed with that capability)
Basically the last NEB wins, but since the same datastructure is passed to all NEBs in the order of load. A NEB could check if the value is already filled and then decide wether to overwrite, extend or leave it.

>     * When we do a reload we don't (have to) clear all memory, so we could probably implement some method for NEB modules to initialize memory on first load which isn't reset. Will probably be a bit messy and that wouldn't work on full restarts of course.
Right, maybe some kind of persistent key/value store would be nice. There is a key/value vector library in naemon already, i will check if its suiteable here.

>     * Another solution could be to store the vault secret in a root-only readable file on disk, I believe one could then make use of `StandardInput=file:/path/to/credentials` in systemd but I have not tested this. Slightly less secure than using user input directly, but saves one from having to do things manually.

That sounds promissing.

nook24:
This sounds like I could use it for #362 
I will definitely take a closer look at this after my vacation :)

sni:
right, i am pretty sure you could implement those uuids with this change.

sni:
here is a example module using this https://github.com/naemon/naemon-vimcrypt-vault-broker

nook24:
Is there a timeline when this will be merged into the main branch?

sni:
there is no timeline, its ready when its ready. There are two things open i'd like to test as mentioned in the first post.

sni:
I'd say this is ready. Anything else is left to be implemented in the broker modules.

jacobbaungard (approved):
Looks fine to me!

PR #367, opened 2021-07-12
Update description of date_format in default naemon.cfg 

nook24:
As the title already says, this is just a little change to update the description of the `date_format` option in the default naemon.cfg


sni (approved):

sni:
thanks

PR #369, opened 2021-10-31
extract environment from command_line and set it directly

sni:
This PR removes exposing environment variables to the process list. The
example command line: TEST=1 FOO=BAR /bin/true
would be executed like: /bin/sh -c 'TEST=1 FOO=BAR /bin/true'
and the environement variables would be visible in ps output.

Instead we extend command line parsing and remove environment variable
definitions and put them into the environment before forking and running
the command.

### DIFF
--- a/lib/runcmd.c
+++ b/lib/runcmd.c
@@ -27,6 +27,7 @@
 #include <sys/time.h>
 #include <sys/resource.h>
 #include <errno.h>
+#include <ctype.h>
 #include "runcmd.h"
 
 
@@ -115,6 +116,8 @@ pid_t runcmd_pid(int fd)
 #define STATE_INDQ  (1 << 3)
 #define STATE_SPECIAL (1 << 4)
 #define STATE_BSLASH (1 << 5)
+#define STATE_INVAR (1 << 6)
+#define STATE_INVAL (1 << 7)
 #define in_quotes (state & (STATE_INSQ | STATE_INDQ))
 #define is_state(s) (state == s)
 #define set_state(s) (state = s)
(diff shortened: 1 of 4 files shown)


lib/runcmd.c
```diff
@@ -115,27 +116,133 @@ pid_t runcmd_pid(int fd)
...
+		case 0:
+			return ret;
+
+		case ' ': case '\t': case '\r': case '\n':
+			if(in_quotes)
+				break;
+			if(is_state(STATE_INVAR)) {
+				/* abort checking for variables, this is something else */
+				env--;
+				i = len;
```
jacobbaungard, line 154: Is this different from just doing a break? We set `i = len`, then go on to the next loop iteration, with the loop condition `i < len` guaranteed to fail?

jacobbaungard (commented):
Looks good in general. Found a potential memory leak, plus:

- The env variable parsing logic is a little confusing to me. Is it possible to add some additional comments that explains the logic a bit clearer? 
- Do we support setting env variables to marcros here (i.e `ENV1=$VAULT_MYSQLPW$` ). Would be good with a testcase that includes a macro.

sni:
>     * The env variable parsing logic is a little confusing to me. Is it possible to add some additional comments that explains the logic a bit clearer?

Sure, i will add a few more comments. It took me a while to figure it out myself :-)

>     * Do we support setting env variables to marcros here (i.e `ENV1=$VAULT_MYSQLPW$` ). Would be good with a testcase that includes a macro.
Naemon macros are already expanded at this point, so this is the final command line already with no more macros in it.

lib/runcmd.c
```diff
@@ -115,27 +116,133 @@ pid_t runcmd_pid(int fd)
...
 	int state, ret = 0;
 	int seen_space = 0;
 	size_t len;
-	char *argz;
+	char *argz, *envz;
 
+	// extract leading environment variables
 	set_state(STATE_NONE);
 	len = strlen(str);
+	envz = malloc(len + 1);
```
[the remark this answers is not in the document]
sni: i simply copied the logic from argz and its freed later on. But i will do some more checking to make sure it doesn't leak memory. Right now it is freed for example here: https://github.com/naemon/naemon-core/pull/369/files#diff-a2ffb0bf5fa5a83674ac68ae810c3c824791c482648915004d205431b43255c9R485

lib/runcmd.c
sni: yes, `break` breaks from the current switch and still finishes the for loop (which appends the current character at the end of the loop). But i could add a simple flag, maybe this would make the code more clear.

sni: Basically, we use "break" to break the switch and add a character and "continue" to do nothing and start over with the next character.

jacobbaungard (approved):
OK nice. This looks sensible to me!

Issue #364, opened 2021-05-31
Adding configuration for start and reload


BaldMansMojo:
Hi,

I added the following to my init script. Should be done for systemd too:

    # Added configuration check. Without the check Naemon crashes and you 
    # have to examin the logfile - Martin Fuerstenau
    
    $exec --allow-root -v $config > /dev/null 2>&1
    retval=$?

    if [ $retval -ne 0 ]
       then
       failure
       echo
       echo "CONFIG ERROR! Reload/restart aborted. Check your Naemon configuration."
       return $retval
    fi

When unfortunately doing a reload without checking configuration first Naemon stops. Then you have to look into the logfile..... . this takes time. If this happens when running Naemon in a cluster (I do it with CentOS) the cluster starts restarting Naemon or switching to another node.

Adding this to reload means that the daemon still runs, nothing will happen and you can fix your configuration.

Martin

PR #370, opened 2021-11-02
Implemented: Adding configuration for start and reload (#364)

ccztux:


jacobbaungard:
Thanks for sending this is. We used to have something similar in the past although for restarts, however it was removed (see references below). The past is messy though, and it seems it was never removed for systemd, even though that was probably intended.

The reason at that time, is that for very large systems, reload/restarts/verify-config can take a significant amount of time. Bundling verify-config with the systemd/service commands makes it hard for a user/other software to make a choice what you want to prioritize. I.e. there is nothing stopping you today from running a verify config yourselves before doing the `systemctl`/`service` call, if that's what you prefer.

I personally don't have a strong opinion on the subject (and we ship our own systemd files anyway), but it's not entirely clear to me what the "correct" / "best" approach is.

References:
https://github.com/naemon/naemon-core/pull/170
https://github.com/naemon/naemon-core/pull/175
https://github.com/naemon/naemon-core/pull/177

tomaszdubiel18:
What's the problem to check the scale, the size of naemon installation and to perform the check at the restart when number of services/hosts/etc is not greater than X?
For me it is very expected feature, missing in Naemon, existing in Nagios. For my installation (2388 services, 169 hosts) config check takes miliseconds.
People make mistakes, forget about something and restarting Naemon with existing errors sometimes can take time to find the cause of it.

tomaszdubiel18:
Maybe just add the parameter in Naemon configuration?
"Check Naemon configuration before the restart = true"
with default "No"

jacobbaungard:
> In OMD we have a special environment variable `CORE_NOVERIFY=yes` which controls the behaviour if config checks before reload/restart/start commands. Maybe something like that would be a good idea for the standard rc files here as well.

I think that sounds like a great solution :+1:

jacobbaungard:
> Maybe just add the parameter in Naemon configuration?
> "Check Naemon configuration before the restart = true"
> with default "No"

I think that's not so easy to implement as we'd need to parse the config file somehow then (i.e in the systemd/service files). It would be much simpler to implement with an environment variable.

nook24:
I think this would be useful for users that are not working with Naemon all day long :)

To be fair, I guess most of us are using some tool to generate the Naemon config files for us. I don't think that @jacobbaungard or @sni are messing around with config files anymore.

I also assume that the _power users_ are using own systemd services anyway. Mine for example does a config check:
```
[Service]
Type=forking
PIDFile=/opt/openitc/nagios/var/nagios.lock
ExecStartPre=/opt/openitc/nagios/bin/naemon -v /opt/openitc/etc/nagios/nagios.cfg
ExecStart=/opt/openitc/nagios/bin/naemon -d /opt/openitc/etc/nagios/nagios.cfg
ExecReload=/bin/kill -HUP $MAINPID
```

Also the config generator is running a validation before reloading Naemon:

Bildschirmfoto 2021-11-03 um 10 33 22

On my test system with 5024 hosts and 93363 services the verification takes 2.5 seconds.

In my opinion a config validation should be enabled by default. Users with such big setups where the verification takes too long have enough experience to disable the verification, or are already using own systemd service configurations anyway.

jacobbaungard:
> In my opinion a config validation should be enabled by default. Users with such big setups where the verification takes too long have enough experience to disable the verification, or are already using own systemd service configurations anyway.

OK for me as long as it's easy to disable with i.e an env var. You have any preference for default behaviour @sni ?

sni:
its fine for me if the default is enabled. As long as its easy to disable.

ccztux:
> 
> 
> > In OMD we have a special environment variable `CORE_NOVERIFY=yes` which controls the behaviour if config checks before reload/restart/start commands. Maybe something like that would be a good idea for the standard rc files here as well.
> 
> I think that sounds like a great solution 👍

That sounds great.

sni:
Any news on this one? And could you please rebase onto the latest upstream? thanks

PR #372, opened 2021-11-08
naemon-core must depend on libnaemon

sni:
which is the case for rpms already. And while on it, line up dependencies.


jacobbaungard (approved):
Looks good to me!

Issue #374, opened 2021-11-24
Very slow start up of Naemon if a system has many downtimes


nook24:
Hi,

I found an issue within Naemon which will occur as soon as the system has to handle thousands of downtimes. In this case a restart of Naemon takes more than 15 minutes.

```
# cat retention.dat | grep hostdowntime | wc -l
21127

# cat retention.dat | grep servicedowntime | wc -l
294420
```

The  issue is caused by the `find_comment` function
https://github.com/naemon/naemon-core/blob/47002195d849455863c2fad7f1a440192ef6cce5/src/naemon/comments.c#L640-L650

My suggestion is to replace this with an hashmap.

nook24:
I guess this will be resolved with https://github.com/naemon/naemon-core/pull/440 ?@sni

sni:
that's right.

PR #375, opened 2021-11-24
Use hasmap to find comment data #374

nook24:
This is my attempt to patch #374. Would be good if you could review this.
Signed-off-by: nook24 <info@nook24.eu>

### DIFF
--- a/src/naemon/comments.c
+++ b/src/naemon/comments.c
@@ -9,20 +9,31 @@
 #include "events.h"
 #include "globals.h"
 #include "nm_alloc.h"
+#include <glib.h>
 
 comment *comment_list = NULL;
 int defer_comment_sorting = 0;
 comment **comment_hashlist = NULL;
+static GHashTable *comment_hashtable;
 
 
 /******************************************************************/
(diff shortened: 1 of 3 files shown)


sni:
Now there is a `comment_hashtable` and a `comment_hashlist`. Have you checked if we need both by any chance?

nook24:
Tbh, I'm not sure if I understand whats happening with `comment_hashlist` by 100%. It is only used by `comment *get_next_comment_by_host(char *host_name, comment *start)`. I think it is possible to get rid of it but I tried to touched as less code as possible.

How ever, by removing `comment_hashlist` we could also delete this:
https://github.com/naemon/naemon-core/blob/47002195d849455863c2fad7f1a440192ef6cce5/src/naemon/shared.c#L416-L437

sni (approved):
looks good

sni:
could you please rebase this PR on the latest HEAD. This should also trigger the checks again.

sni:
i made some changes and created https://github.com/naemon/naemon-core/pull/420

PR #376, opened 2021-11-29
pass commands containing tilde to shell

sni:
Commands containing a tilde ~ should be passed to the shell. execvp does not expand it.


jacobbaungard (approved):
Looks good to me

PR #379, opened 2022-01-28
Cast time_t variables so tests run on Debian

lamaral:
While running the tests on Debian, they were failing due to an incompatibility on the size of time_t and the format that was being used when printing the messages in the tests. 

Casting the variables fixed the tests on Debian. After the changes, I ran the tests inside a Ubuntu 20.04 container to make sure they didn't break and everything was fine.


sni (approved):
looks good, thanks

sni:
i think i ran into this as well some time ago

PR #382, opened 2022-03-10
let all links point to https://www.naemon.io

sni:


jacobbaungard (approved):

nook24:
Will be naemon.org still available or should we all adjust our links?

nook24:
Looks like naemon.org is dead now. This was quite fast. The Naemon's GitHub profile is also still using the .org domain:
Bildschirmfoto 2022-03-23 um 08 56 59

sni:
thanks, i fixed that link. naemon.org is dead, that's right and that's the reason for this move. What's still working is www.naemon.org.

PR #383, opened 2022-03-16
set last_update broker_*_status

sni:
the broker status is usually called when the object changes, so lets update
the last_update attribute as well.
This addresses an issue where last_update does not get updated when a downtime ends.


jacobbaungard (approved):
Sure, makes sense :+1:

PR #389, opened 2022-07-21
debian: change libdir to standard location

sni:
on debian/ubuntu with multiarch packages must not install libraries in `/usr/lib` but in `/usr/lib/<arch>`

### DIFF
--- a/debian/libnaemon.install
+++ b/debian/libnaemon.install
@@ -1,1 +1,0 @@
-/usr/lib/naemon/libnaemon.so*
--- a/debian/naemon-dev.install
+++ b/debian/naemon-dev.install
@@ -1,3 +1,1 @@
 /usr/include/naemon
-/usr/lib/naemon/libnaemon.a
-/usr/lib/naemon/libnaemon.la
--- a/debian/rules
+++ b/debian/rules
@@ -5,13 +5,18 @@
 
 export DH_VERBOSE=1
 DESTDIR=$(CURDIR)/debian/tmp/
+ifeq ($(DEB_HOST_MULTIARCH),)
+	LIBDIR=/usr/lib/naemon
+else
+	LIBDIR=/usr/lib/$(DEB_HOST_MULTIARCH)
+endif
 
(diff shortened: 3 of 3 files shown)


roelvanmeer:
Hi Sven, I think this can be simplified quite a bit. I don't think you need to separately handle the case where DEB_HOST_MULTIARCH is empty. Also, dynamically adding lines to the install file isn't too pretty. If, after building, the files are installed in the correct location, the install file could contain a line like this:

`usr/lib/*   /usr/lib`

and files will be installed correctly, whether they are in usr/lib/naemon or in /usr/lib/x86_64-linux-gnu/naemon

https://wiki.debian.org/Multiarch/Implementation#dh.281.29_and_autotools gives some hints on what to do, including setting DEB_HOST_MULTIARCH in d/rules, which isn't done now.

I've never converted a package to multiarch before, so this is just theoretical knowledge, but if you like I can have a go at this. I'm always willing to up my Debian packaging skills.

sni:
well, that would be great if course. I don't think we need to support both ways. A quick search showed it is supported at least since ubuntu 12. So i guess it's fine :-)

roelvanmeer:
All right, after working on this a bit I have a few quick questions. :)

First, is there a reason that libnaemon is installed in /usr/lib/naemon and not in /usr/lib? The related question of course is whether we install libnaemon in /usr/lib/$DEB_HOST_MULTIARCH/, or in /usr/lib/$DEB_HOST_MULTIARCH/naemon/? I think the first is the more common case, but that's not how it is, currently.

The second question is about how far you're willing to take this. The library package should be called libnaemon0 if you really want to follow the Debian standards, but that might be taking it too far for now..

sni:
No, there is no specific reason to do so. It seemed to be the right place i guess. Personally i would like to follow the debian standards unless there are good reasons not to do so.
But changing this might break existing neb module addons, that's the one thing we should check. For naemon itself it shouldn't be an issue.

sni:
obsolete by #393

PR #390, opened 2022-08-16
Debian packaging improvements

roelvanmeer:


sni:
Looks good, thanks. Except, did you miss to `git add debian/naemon-core.logrotate`. It looks like that file is missing.

roelvanmeer:
Ah, good one! I see it's mentioned in .gitignore, that's why I missed it. Should be correct now.

sni:
right, since the file was created during build time, we put it into the gitignore. If it is a regular file now, it should be removed from the gitignore of course.

roelvanmeer:
> If it is a regular file now, it should be removed from the gitignore of course.

No it isn't. I just thought it was already in the repo because it didn't show up in my git status. We're still using the copy of naemon.logrotate.el7. Essentially the second commit now only stops installing the built logrotate file, which was overwritten by the one installed by installlogrotate anyway.

roelvanmeer:
This might break the build on Debian jessie, now that I think about it, but support for that version ended two years ago.

sni:
i guess that's fine. Thanks

sni (approved):

PR #396, opened 2022-09-19
fix CHANGE_HOST_CHECK_TIMEPERIOD command

sni:
command crashes because timeperiod is stored with different name.


jacobbaungard (approved):
Nice, looks good. A test could be nice perhaps if you have time, otherwise OK :+1:

sni:
here comes the test case :-)

PR #402, opened 2022-11-18
close worker standard filehandles (fixes #401)

### LINKED ISSUE #401: naemon doesn't properly close stdout/stderr/stdin
Hi,

i think the faulty commit is fa9c19dbf8cc37e3d81b8bb223a81e08ef4bc151.

there are some close(0) .. close(2) lines, they where moved from daemon_init to main. daemon_init is called before init_worker but now the fds are closed after init_worker, so they are only closed in the main process and not in the workers. That leads to the following problem:

code like this does not return:

``` shell
output=$( ./src/naemon/naemon  -du sample-config/naemon.cfg )
```

in reality i'm doing `output=$( omd restart )` and that doesn't work anymore.

Before this commit it works as expected.

So I think in init_worker you also need to close the fds?

sni:


jacobbaungard (approved):
Looks reasonable to me :+1:

PR #403, opened 2022-11-19
packaging: remove non-existing configure options

sni:
Those config options don't do anything and configure complains they do not exist.


jacobbaungard (approved):
I was kinda surprised about `with-naemon-config-dir`. We're definitely using that in other projects like livestatus, in order to define where the NEB module config file goes, but indeed it doesn't look like it is used in the core (thought perhaps it would put it in pkgconfig or so).

Anyway, I guess it's not used, so all good :+1:

sni:
Actually i was surprised as well and considered adding the option. But then again, it does not make sense for naemon-core to add a config option for each config subfolder.
It makes more sense for livestatus, where this option is the main folder for its configuration.

PR #404, opened 2022-11-25
Fix memory leak in cmd parser

sni:
Fixing two issues here. If you like separate PRs, just tell me :-)

First one fixes a memory leak in the cmd parser which makes the worker processes grow which
then leads to more cpu usage during all the forking (along with the wasted memory)

The second came to my attention while searching the leak, those two allocations where by far the biggest ones on my test system and both simply depend on the maximum number of open files.
I added a high enough max value to be hopefully future proof without wasting useless memory.
I don't think any naemon worker will ever sanely use more than 100k open file handles. I 
considered adding a log entry or another warning. But at that point, there is no access to the log and stderr/stdout is already closed.
With those patches, memory consumtion seems stable.

### DIFF
--- a/lib/iobroker.c
+++ b/lib/iobroker.c
@@ -140,6 +140,10 @@ struct iobroker_set *iobroker_create(void)
 	}
 
 	iobs->max_fds = iobroker_max_usable_fds();
+	/* add sane max limit, if ulimit is set to unlimited or a very high value we
+	 * don't want to waste memory for nothing */
+	if (iobs->max_fds > MAX_FD_LIMIT)
+		iobs->max_fds = MAX_FD_LIMIT;
 	iobs->iobroker_fds = calloc(iobs->max_fds, sizeof(iobroker_fd *));
 	if (!iobs->iobroker_fds) {
 		goto error_out;
--- a/lib/runcmd.c
+++ b/lib/runcmd.c
@@ -69,7 +69,6 @@ static int maxfd = 0;
 # endif /* _SC_OPEN_MAX */
 #endif /* OPEN_MAX */
 
-
 const char *runcmd_strerror(int code)
 {
 	switch (code) {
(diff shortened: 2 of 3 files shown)


lib/iobroker.c
```diff
@@ -140,6 +140,10 @@ struct iobroker_set *iobroker_create(void)
 	}
 
 	iobs->max_fds = iobroker_max_usable_fds();
+	/* add sane max limit, if ulimit is set to unlimited or a very high value we
+	 * don't want to waste memory for nothing */
+	if (iobs->max_fds > 100000)
+		iobs->max_fds = 100000;
```
jacobbaungard, line 146: Could we make this configurable just in case someone wants a different value?

lib/runcmd.c
```diff
@@ -140,6 +140,7 @@ int runcmd_cmd2strv(const char *str, int *out_argc, char **out_argv, int *out_en
 	len = strlen(str);
 	envz = malloc(len + 1);
 	*out_envc = env;
+	out_env[0] = envz;
```
jacobbaungard, line 143: Is t hat the one free `free` later with `free(env[0]);`  ?

lib/iobroker.c
sni: i thought about it as well. It is usually limited by `ulimit -n`. So if you want a smaller value, set a sane ulimit.
Not sure if a config option is required, maybe a compile time constant?

jacobbaungard: Yeah I guess it is mainly for larger values on would nee to change it. 

There shouldn't be much need to increase it above this limit I suppose, but a configuration option somewhere (compiler/runtime) seems nicer than a "magic" constant.

sni:
i removed the magic... :-)

jacobbaungard (approved):
Great, thanks!

PR #405, opened 2022-11-30
rpm: fix fg: no job control

sni:
using an undefined macro trys to run literal %{insserv_cleanup} as shell command which results in the errors like

    /var/tmp/rpm-tmp.1P1HM5: line 12: fg: no job control


jacobbaungard (approved):
Looks good

Issue #409, opened 2022-12-30
systemd-tmpfiles-clean.service reports legacy directory /var/run/naemon


midnight-run:
Hello

this should be changed in a future version of naemon-core:
systemd-tmpfiles[2111098]: [/usr/lib/tmpfiles.d/naemon.conf:1] Line references path below legacy directory /var/run/, updating /var/run/naemon → /run/naemon; please update the tmpfiles.d/ drop-in file accordingly.

naemon-core version 1.4 on Red Hat Enterprise Linux 8/(9) (Systemd version systemd-239 and above)

$ systemctl restart systemd-tmpfiles-clean.service
$ tail messages
daemon.info systemd[1]: Stopped User Manager for UID 0.
daemon.info systemd[1]: Stopping User runtime directory /run/user/0...
daemon.info systemd[1]: run-user-0.mount: Succeeded.
daemon.info systemd[1]: user-runtime-dir@0.service: Succeeded.
daemon.info systemd[1]: Stopped User runtime directory /run/user/0.
daemon.info systemd[1]: Removed slice User Slice of UID 0.
daemon.info systemd[1]: Starting Cleanup of Temporary Directories...
**daemon.notice systemd-tmpfiles[2111098]: [/usr/lib/tmpfiles.d/naemon.conf:1] Line references path below legacy directory /var/run/, updating /var/run/naemon → /run/naemon; please update the tmpfiles.d/ drop-in file accordingly.**
daemon.info systemd[1]: systemd-tmpfiles-clean.service: Succeeded.
daemon.info systemd[1]: Started Cleanup of Temporary Directories.

$ cat /usr/lib/tmpfiles.d/naemon.conf
D /var/run/naemon 0755 naemon naemon -
r /var/cache/naemon/checkresults/* 0644 naemon naemon 1d

midnight-run:
yes it works for rhel7 as well.

- Please add following to postinstall scriptlet in naemon-core.spec (at the end of %post)
systemd-tmpfiles --create %{_tmpfilesdir}/naemon.conf

- Please remove this 2 lines in daemon-systemd.in (not needed)
ExecStartPre=-/usr/bin/mkdir -p /var/run/naemon
ExecStartPre=/usr/bin/chown -R naemon:naemon /var/run/naemon/

midnight-run:
and set pidfile lockfile to /run/naemon/naemon.pid instead of /var/run/naemon/naemon.pid (definition for @lockfile@)

[...]

PR #410, opened 2023-01-11
fix systemd configuration from using legacy directory and apply systemd-tmpfiles settings

midnight-run:
This is my attempt to patch #409 . Please verify my changes.
Signed-off-by: midnight-run 1+dasgithub@kmdinata.ch


sni (approved):

sni:
Great. Thanks. I tested the build on rhel7, rhel8 and rhel9 and it seems fine.

PR #411, opened 2023-01-12
add if around systemd call

sni:
add a if clause around the systemd-tmpfiles call like we do with all the other systemd calls.


sni:
this adds a missing if introduced by #410.

jacobbaungard (approved):
Jep makes sense

PR #415, opened 2023-02-03
Support el9

jacobbaungard:
- Add conditionals for EL9 support.
- Use the pre-defined `optflags` variable to set the CFLAGS.

### DIFF
--- a/naemon-core.spec
+++ b/naemon-core.spec
@@ -1,12 +1,5 @@
 %define logmsg logger -t naemon/rpm
 
-# Setup some debugging options in case we build with --with debug
-%if %{defined _with_debug}
-  %define mycflags -O0 -pg -ggdb3
-%else
-  %define mycflags %{nil}
-%endif
-
 Summary: Open Source Host, Service And Network Monitoring Program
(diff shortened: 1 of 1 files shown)


naemon-core.spec
```diff
@@ -4,7 +4,11 @@
 %if %{defined _with_debug}
   %define mycflags -O0 -pg -ggdb3
 %else
-  %define mycflags %{nil}
+  %if %{defined optflags}
+    %define mycflags %{optflags}
```
jacobbaungard, line 8: I'm not 100% sure how to verify that this is actually used, but hopefully it works. If any idea how we can checked what cflags are actually used at compiletime.

sni: tbh, i would just remove all this mycflags. We provide standard packages. If anyone requires something specific, they have to build their own anyway.

sni (approved):

naemon-core.spec
jacobbaungard: Ok, yeah we can do that, then the standard CFLAGS env variable should just be used I would think.

PR #420, opened 2023-03-01
Rework comment handling

sni:
rework comment handling
    
This PR reworks the comment handling to improve performance when having lots of comments.
    
  - it add a prev pointer to the comments list so we don't have to iterate over the complete list when removing an item
  - it replaces the hash lookup to quickly access comments for specific hosts/services in favor of a new host/service attribute which holds links to the comments
  - it avoids duplicate host/service lookups for DEL_ALL_HOST_COMMENTS/DEL_ALL_SVC_COMMENTS command
  - it removes hash functions which are now unused
    
In general, it tries to adopt the downtime handling over to comments.
    
This PR is based on and includes #375.
    
Since the internal object structure changes, all NEB modules have to be rebuild. (We therefor increase CURRENT_NEB_API_VERSION to 6)
    
Signed-off-by: Sven Nierlein <sven@consol.de>

### DIFF
--- a/src/naemon/comments.c
+++ b/src/naemon/comments.c
@@ -9,20 +9,31 @@
 #include "events.h"
 #include "globals.h"
 #include "nm_alloc.h"
+#include <glib.h>
 
 comment *comment_list = NULL;
 int defer_comment_sorting = 0;
 comment **comment_hashlist = NULL;
+static GHashTable *comment_hashtable;
 
 
 /******************************************************************/
 /**************** INITIALIZATION/CLEANUP FUNCTIONS ****************/
 /******************************************************************/
 
+int initialize_comment_hashmap(void)
+{
+	if (comment_hashtable == NULL){
+		comment_hashtable = g_hash_table_new(g_direct_hash, g_direct_equal);
+	}
+	return OK;
+}
+
 /* initializes comment data */
 int initialize_comment_data(void)
 {
 	comment *temp_comment = NULL;
+	initialize_comment_hashmap();
 
 	/* find the new starting index for comment id if its missing*/
 	if (next_comment_id == 0L) {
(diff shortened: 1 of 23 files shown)


jdumalaonITRS:
Hi, 
Not sure if I missed something but doesn't it look like we need the comment_list even less? 
We can still have the host/service attribute which holds links to the comments and have a hashtable. 

I am just not 100% sure what needs to happen for the sorting code for comment_list though. Does anyone know why we need to sort?

sni:
I assumed the comment_list is required for livestatus, but i did a quick search and did not find any reference. It seems like comments are soley add/removed from broker callbacks.
I can try to remove it completely. And if it's removed, we don't have to order it.

sni:
Alright, i completely removed the comments_list now.

jdumalaonITRS:
Nice thanks for this. 
The code looks good. Let me try some manual tests for this. I'll try to feedback by next week.

jdumalaonITRS (approved):
LGTM. Thanks!

src/naemon/comments.c
```diff
@@ -142,52 +136,27 @@ int add_new_service_comment(int entry_type, char *host_name, char *svc_descripti
...
-	if (comment_list == this_comment)
-		comment_list = this_comment->next;
-	else
-		last_comment->next = next_comment;
+	// remove from svc or host
+	if (type == HOST_COMMENT) {
+		host *temp_host = find_host(this_comment->host_name);
+		remove_object_from_objectlist(&temp_host->comments_list, this_comment);
+	}
+	else if (type == SERVICE_COMMENT) {
```
nook24, line 155: Personally i would prefer a simple `else` statement here

nook24 (approved):

nook24:
I have tested this with 400k comments, works as promised.
````
[0.0079 (+0.0000)] Initialized retention data
[0.0079 (+0.0000)] Reading initial state information
[2.4028 (+2.3948)] Read initial state information
[2.4029 (+0.0002)] Restored 0 downtimes
[2.4029 (+0.0000)] Restored 400000 comments
[2.4030 (+0.0000)] Initializing performance data
````

PR #422, opened 2023-03-04
improve latency handling

sni:
right now all hosts/services have a wrong latency because calculation is based on the integer value of next_check but the scheduled event uses microseconds. So use the latency based on the event timestamp instead which has subsecond precision.

### DIFF
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -143,7 +143,6 @@ static void handle_service_check_event(struct nm_event_execution_properties *evp
 	int nudge_seconds = 0;
 	double latency;
 	struct timeval tv;
-	struct timeval event_runtime;
 	int options = temp_service->check_options;
 	host *temp_host = NULL;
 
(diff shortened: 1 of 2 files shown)


src/naemon/checks_service.c
```diff
@@ -486,6 +483,8 @@ int handle_async_service_check_result(service *temp_service, check_result *queue
 	/* was this check passive or active? */
 	temp_service->check_type = (queued_check_result->check_type == CHECK_TYPE_ACTIVE) ? CHECK_TYPE_ACTIVE : CHECK_TYPE_PASSIVE;
 
+	temp_service->latency = queued_check_result->latency;
```
jacobbaungard, line 486: Why is this one needed? Should there be a similar place where the same is required for hosts?

sni: You are right, but for hosts this is done already here https://github.com/naemon/naemon-core/blob/master/src/naemon/checks_host.c#L407

Thing is, this does not make a huge difference for naemon-core alone because in naemon, the latency at the start of the check is the final latency already. But mod-gearman set the latency into the check-result when it's finished running a job and this latency can be higher.

jacobbaungard: Right. Merlin probably does something similar, but can't remember. This should be fine for the core anyway.

jacobbaungard (approved):

PR #431, opened 2023-06-28
use dynamic sized command line buffer

sni:
Previously we used a fixed size 8k buffer when parsing command line arguments. This sounds much, but there are command lines bigger than 8k and they are simply cut off without any warning.

Instead we use a dynamic sized buffer with the size of the raw command now.

(while on it, fixing the crazy indent)

### DIFF
--- a/src/naemon/macros.c
+++ b/src/naemon/macros.c
@@ -121,8 +121,9 @@ static int grab_custom_object_macro_r(nagios_macros *mac, char *macro_name, cust
 /* given a "raw" command, return the "expanded" or "whole" command line */
 int get_raw_command_line_r(nagios_macros *mac, command *cmd_ptr, char *cmd, char **full_command, int macro_options)
 {
-	char temp_arg[MAX_COMMAND_BUFFER] = "";
+	char *temp_arg = NULL;
 	char *arg_buffer = NULL;
+	size_t cmd_len = 0;
 	register int x = 0;
 	register int y = 0;
 	register int arg_index = 0;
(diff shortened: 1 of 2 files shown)


src/naemon/macros.c
```diff
@@ -139,51 +140,56 @@ int get_raw_command_line_r(nagios_macros *mac, command *cmd_ptr, char *cmd, char
 	/* get the full command line */
 	*full_command = nm_strdup((cmd_ptr->command_line == NULL) ? "" : cmd_ptr->command_line);
 
-	/* XXX: Crazy indent */
-	/* get the command arguments */
-	if (cmd != NULL) {
+	if (cmd == NULL) {
+		log_debug_info(DEBUGL_COMMANDS | DEBUGL_CHECKS | DEBUGL_MACROS, 2, "Expanded Command Output: %s\n", *full_command);
```
jacobbaungard, line 144: In this case, it means that there was no need to further expand the command? I.e for commands like `DISABLE_NOTIFICATIONS` where there are no args

jacobbaungard: sorry, this is not external commands, but check commands... Similar point though.

sni: tbh, i didn't check where this is used, i just inverted the huge if clause to return early instead of having the hole function in a huge if. The functionality did not change at this point.
I should have put this in a separate commit.

jacobbaungard (commented):
My C is probably already rusty, but don't we just put a single, un-expanded check argument into the `temp_arg` variable? 

I.e if we have `example_command!$_MY_MACRO$` we end up putting `$_MY_MACRO$` into `temp_arg` and passing it to `process_macros_r` which puts the expanded version of the macro in `arg_buffer` ?

sni:
that's right, but in my case i had command with very long arguments like `example_command!some very long arg!other arg` and it was cut off, if the argument was more than 8k.

jacobbaungard:
You had a single argument that was above 8k bytes!?

sni:
right, basically a tarball of a folder as base64 encoded string :-) I know, not the usual usecase, but naemon should either work or throw an error :-)

jacobbaungard (approved):
Jeez OK! Seems sensible enough. I guess we might even save some bytes on normal smaller commands like this.

When I first read the title I figured it would be the total check_command (expanded/non expanded) that we change, so I guess just make sure to note in the changelog that it's the size of the individual command line args that can now be longer.

nook24:
> When I first read the title I figured it would be the total check_command (expanded/non expanded) that we change, so I guess just make sure to note in the changelog that it's the size of the individual command line args that can now be longer.

Honestly, I was also testing the total check command length first^^

sni:
sorry for the confusion :-)
i will try to make the changelog entry more clear.

thanks for the review.

PR #438, opened 2023-09-17
add systemd to build requires

sni:
this (should) fix this build error on obs:
```
[  139s] /.build_patchrpmcheck_scr: line 55: systemd-tmpfiles: command not found
[  139s] postinstall script of naemon-core-1.4.1-lp154.18.1.x86_64.rpm failed
```

### DIFF
--- a/naemon-core.spec
+++ b/naemon-core.spec
@@ -41,6 +41,7 @@ Requires(pre): systemd
 Requires(post): systemd
 Requires(preun): systemd
 Requires(postun): systemd
+BuildRequires: pkgconfig(systemd)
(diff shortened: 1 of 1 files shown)


nook24:
I have no idea about rpm spec files. With that said, the `systemd` package contains the `systemd-tmpfiles` binary on all distributions :)

According to the Fedora docs, the syntax is correct, so I'm going to approve this
https://docs.fedoraproject.org/en-US/packaging-guidelines/PkgConfigBuildRequires/

sni:
Thanks,
for now, i've added it in the suse specific block. So it should not affect fedora or rhel builds anyway.

PR #439, opened 2023-10-12
write objects.precache in tmp file first, then move it

sni:
It's a good standard to do so and in fact, we do this already in several places, ex. the status.dat. This ensures the file is ready and completly written before it will be used.

The issue here is, that naemon starts without any issues if the precached file is empty for any reason. Except it has zero hosts/services then and removes all existing states/downtime/comments.

### DIFF
--- a/src/naemon/objects.c
+++ b/src/naemon/objects.c
@@ -11,6 +11,7 @@
 #include "logging.h"
 #include "globals.h"
 #include "nm_alloc.h"
+#include "utils.h"
 
 int __nagios_object_structure_version = CURRENT_OBJECT_STRUCTURE_VERSION;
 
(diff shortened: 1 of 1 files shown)


src/naemon/objects.c
```diff
@@ -109,7 +124,41 @@ int fcache_objects(char *cache_file)
...
+
+	/* close the temp file */
+	result = ferror(fp) | fclose(fp);
+
+	/* save/close was successful */
+	if (result == 0) {
+
+		result = OK;
+
+		/* move the temp file to the status log (overwrite the old status log) */
```
jdumalaonITRS, line 144: Just a minor comment here. I guess status log should be cache file.

jdumalaonITRS (approved):
LGTM. Just a minor comment in the code comment

src/naemon/objects.c
sni: indeed, missed that while copy/pasting from the status.dat code.

sni: should be fine now

PR #441, opened 2024-01-15
Define ETIME for systems without that constant

dereckson:
ETIME alternative implementation as ETIMEDOUT is already available for the worker component, but not for the naemon component.

This fixes the following compilation error on systems without ETIME:

```
src/naemon/workers.c:517:27: error: use of undeclared identifier 'ETIME'
                if (wpres.error_code == ETIME) {
                                        ^
1 error generated.
```

This has been noticed on FreeBSD.

### DIFF
--- a/src/naemon/workers.h
+++ b/src/naemon/workers.h
@@ -12,6 +12,10 @@
 
 #define WPROC_FORCE  (1 << 0)
 
+#ifndef ETIME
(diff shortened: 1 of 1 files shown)


sni (approved):

sni:
thanks for your contribution

PR #442, opened 2024-01-15
Include <sys/wait.h> for wait constants

dereckson:
Constants like WIFEXITED are defined in sys/wait.h header.

This fixes a compilation issue on FreeBSD where those contants aren't implicitly made available without an explicit include of those.


sni (approved):

sni:
looks good, thanks

PR #447, opened 2024-01-25
add problem timestamps and duration

sni:
this PR makes hosts / services save the start and end timestamp of the current problem. Those values can then be used as macros, ex. in notification scripts. For this, there are several new macros available:

- `$HOSTPROBLEMSTART$` start timestamp of problem
- `$HOSTPROBLEMEND$` end timestamp of problem (or zero if problem still persists)
- `$HOSTPROBLEMDURATIONSEC$` duration of problem
- `$HOSTPROBLEMDURATION$` duration as human readable text

the same macros exist for services:
- `$SERVICEPROBLEMSTART$`
- `$SERVICEPROBLEMEND$`
- `$SERVICEPROBLEMDURATIONSEC$`
- `$SERVICEPROBLEMDURATION$`

While there is a currently ongoing problem, the values point to this current problem and the end timestamp is zero. Once the problem is resolved, the values can still be used and won't change until a new problem starts.

This makes it possible to use the problem duration in recovery notifications which otherwise would not be possible.

Since this change affects the host/service structs, the neb api version has to be increased and NEB modules have to be rebuild. Once this is accepted, i will add PRs to add those columns to livestatus and the docs on the webpage.

### DIFF
--- a/lib/nsutils.c
+++ b/lib/nsutils.c
@@ -82,7 +82,19 @@ const char *mkstr(const char *fmt, ...)
 	return ret;
 }
 
+/* format duration seconds into human readable string */
+const char* duration_string(unsigned long duration) {
+	int days, hours, minutes, seconds;
 
+	days = duration / 86400;
+	duration -= (days * 86400);
+	hours = duration / 3600;
+	duration -= (hours * 3600);
+	minutes = duration / 60;
+	duration -= (minutes * 60);
+	seconds = duration;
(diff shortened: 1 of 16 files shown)


sni:
Adding the uuid commit here, since it overlaps a bit with the problem macros. If that's an issue, i can move it into a separate PR.

nook24 (approved):

sni:
thanks for the review

PR #449, opened 2024-02-02
fix resetting modified_attributes

sni:
resetting modified_attributes to zero had no effect. The reason was a wrong assumption when reading the retention file. The retained value was restored even if the correspondig modified_attributes was not set (anymore).

Now we have the following attribute changes (assuming retention is enabled):

- attribute was changed via gui -> retained on next reload
- attribute was changed via gui but config value changed afterwards -> use the configured value instead of the retained one
- config value changed and attribute has not been changed via gui -> use the configured value


nook24 (approved):
🦄

Issue #454, opened 2024-02-20
Naemon service fails to start after 1.4.2-19.1 RPM upgrade


imre-sidn:
We have a dnf-automatic timer that periodically updates our RPM packages. Last night a number of packages, including Naemon, got updated. Unfortunately the naemon.service systemd unit got restarted as part of this update, and did not come up correctly.

Relevant log excerpts:

/var/log/dnf/dnf.log
```
2024-02-20T04:34:26+0000 DEBUG Upgraded: gnutls-3.6.16-8.el8_9.1.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: libmaxminddb-1.2.0-10.el8_9.1.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: libnaemon-1.4.2-19.1.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: libssh-0.9.6-13.el8_9.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: libssh-config-0.9.6-13.el8_9.noarch
2024-02-20T04:34:26+0000 DEBUG Upgraded: naemon-1.4.2-14.1.noarch
2024-02-20T04:34:26+0000 DEBUG Upgraded: naemon-core-1.4.2-19.1.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: naemon-livestatus-1.4.2-15.1.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: naemon-thruk-1.4.2-14.1.noarch
2024-02-20T04:34:26+0000 DEBUG Upgraded: naemon-vimvault-1.4.2-4.1.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: nss-3.90.0-6.el8_9.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: nss-softokn-3.90.0-6.el8_9.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: nss-softokn-freebl-3.90.0-6.el8_9.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: nss-sysinit-3.90.0-6.el8_9.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: nss-tools-3.90.0-6.el8_9.x86_64
2024-02-20T04:34:26+0000 DEBUG Upgraded: nss-util-3.90.0-6.el8_9.x86_64
[...]

PR #456, opened 2024-03-01
fix(rpm): Delay Naemon restart until posttrans

imre-sidn:
The posttrans scriptlet executes after the RPM transaction has concluded, i.e. when all packages have been installed or upgraded. Delaying the Naemon restart until posttrans prevents API compatibility issues when naemon-core upgrades before naemon-livestatus. See https://github.com/naemon/naemon-core/issues/454.

Note that it is not possible to infer the RPM operation (install or upgrade) from the posttrans scriptlet argument, since they are the same: https://en.opensuse.org/openSUSE:Packaging_scriptlet_snippets#Syntax This should not cause any problems since the scriptlet only restarts Naemon if it was already running. In other words, it won't cause Naemon to start immediately after installation.

### DIFF
--- a/naemon-core.spec
+++ b/naemon-core.spec
@@ -195,16 +195,8 @@ fi
 %post
 case "$*" in
   2)
-    # Upgrading so try and restart if already running
-    # For systemctl systems we need to reload the configs
-    # becaues it'll complain if we just installed a new
-    # init script
-    %if %{?_unitdir:1}0
-      systemctl daemon-reload &>/dev/null || true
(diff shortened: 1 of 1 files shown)


imre-sidn:
I've tested this change in a Rocky Linux 8 container, by upgrading from the previous 1.4.1 version. I have also tested RPM installs.

sni (approved):

sni:
thanks for digging into and fixing this. This comes just in time, the next release will come
with rebuild neb modules again.

PR #459, opened 2024-03-18
add new options to keep services running as long as they are ok

sni:
The issue with the options:

- host_down_disable_service_checks
- service_skip_check_dependency_status
- service_skip_check_host_down_status
- host_skip_check_dependency_status

is that reports break because hosts/services suddenly stop executing and
keep their OK state. Which makes those options pretty unusable.

So in order to keep reporting correct, you need to keep services running, even
if the host is down. With these new options, hosts/services keep on running as
long as they are up. And as soon as the service is down, it stops running until
the host comes back up. That way naemon has to do less checks, especially less
checks which run into timeouts and such but reporting is still correct.

The option service_skip_check_dependency_status=-2 will also be used for service parents.

Adding a new option service_parents_disable_service_checks to prevent running service
checks if service parents are down.

Recommended settings are:

    host_down_disable_service_checks=1       ; disable service checks if host is down
    service_parents_disable_service_checks=1 ; also disable service checks if parents are down
    service_skip_check_host_down_status=-2   ; but keep running as long as they are ok
    service_skip_check_dependency_status=-2  ; same, but for dependency checks.
    host_skip_check_dependency_status=-2     ; and for host checks.

### DIFF
--- a/sample-config/naemon.cfg.in
+++ b/sample-config/naemon.cfg.in
@@ -1041,18 +1041,27 @@ allow_empty_hostgroup_assignment=0
 # This option will disable all service checks if the host is not in an UP state
 #
 # While desirable in some environments, enabling this value can distort report
-# values as the expected quantity of checks will not have been performed
-
+# values as the expected quantity of checks will not have been performed.
+# Set service_skip_check_host_down_status to -2 to mitigate this.
 #host_down_disable_service_checks=0
+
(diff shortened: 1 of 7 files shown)


sni:
PR is ready and runs here since 2 days without issues.

nook24 (approved):
:)

PR #463, opened 2024-05-20
Add display_status property to hosts and services objects

dirtyren:
This is a idea we have been using for quite sometime in order to gain performance when filtering and adding a new concept of status to naemon.

   The idea is to order items by  priority and as shown in the image bellow.

status_colors

Before I fix any of the required steps to merge, let me see what you guys think about it.

Tks.


sni:
tbh, this is a hard coded and very opinionated sort order. I don't see a general benefit.
Could you explain how this improves performance?

dirtyren:
For this specific sort, it increases performance when doing in livestatus. 
This is  just a proposal to order objects based on incidents that should be checked first in a monitoring system, in my opinion, off course. And also the ideia for the new statuses colors I find it useful.

Tks anyway, we can close this pull request, no problem.

PR #464, opened 2024-05-28
Add option -i to ignore config warnings.

johanthoren:
This is an attempt to improve reload speed in huge environments. If Warnings are not going to be acted upon, an option to skip checking for them during the pre-flight-check or when running a verification would save some seconds for really large environments.

- Add option -i / --ignore-warnings, which will skip checking for configuration issues that only yield warnings. Errors will still be looked for.

### DIFF
--- a/src/naemon/configuration.c
+++ b/src/naemon/configuration.c
@@ -1451,13 +1451,13 @@ int pre_flight_object_check(int *w, int *e)
 	/*****************************************/
 	/* check each service...                 */
 	/*****************************************/
-	total_objects = 0;
-	for (temp_service = service_list; temp_service != NULL; temp_service = temp_service->next) {
+	/* Since these checks only generate warnings, skip them all if warnings are ignored. */
+	if (!ignore_warnings) {
+		total_objects = 0;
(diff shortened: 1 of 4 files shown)


nook24:
Hi @johanthoren, thanks for your contribution!
Could you please provide some numbers about how many hosts and services you have and how much this change improved the restart performance?

Also the output of a Naemon startup with `--enable-timing-point` would be interesting.

johanthoren:
> Hi @johanthoren, thanks for your contribution! Could you please provide some numbers about how many hosts and services you have and how much this change improved the restart performance?
> 
> Also the output of a Naemon startup with `--enable-timing-point` would be interesting.

I will report back once I've convinced the customer to run this patch in the environment where we're trying to reduce the time it takes to reload, but it's likely going to be a few days or more before I can do so since it's in production.

johanthoren:
The change didn't bring significant enough changes. To completely get rid of warnings filling up the logs changes would have to be made in more places than I anticipated, which might be considered a bit too much. So I'll close this pull request.

PR #472, opened 2024-08-06
Fixed missing OK HARD state after OK SOFT.

ccztux:
This pull request fixes the missing `OK HARD` state after an `OK SOFT` state:

image

### DIFF
--- a/src/naemon/checks_service.c
+++ b/src/naemon/checks_service.c
@@ -840,6 +840,9 @@ int handle_async_service_check_result(service *temp_service, check_result *queue
 			/* this is a soft recovery */
 			temp_service->state_type = SOFT_STATE;
 
+			/* this is a soft recovery */
+			temp_service->last_state_was_soft_recovery = TRUE;
+
 			/* log the soft recovery */
 			log_service_event(temp_service);
 			alert_recorded = NEBATTR_CHECK_ALERT;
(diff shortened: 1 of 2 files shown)


sni:
Thanks for your pull request. I'd say what's wrong in the image is the soft recovery at 11:39:29. Which is actually a hard recovery. The soft recovery at 11:39:41 is right.
Also i don't think it is necessary to introduce a new flag for this case.

ccztux:
Apologies for the delayed response. The aim of this pull request was to resolve the issue of missing log entries for OK HARD states that are expected after a soft recovery. I didn’t notice the host’s soft recovery at all.
According to the documentation, the host entered a SOFT UP state at 11:39:41, which is considered OK. However, similar to the behavior observed with services, the corresponding HARD UP state for the host appears to be missing from the log file.
The screenshot illustrates that, as a result of the changes in this pull request, the HARD OK state at 11:44:41 is now properly logged following the SOFT OK state at 11:39:41, which reflects the expected behavior.

ccztux:
Thanks for the feedback! I'm currently working on a solution that avoids introducing a new flag.
While working on these changes, I encountered a specific situation where I'm unsure how Naemon should interpret this:

image

ccztux:
Thank you for your prompt reply.

PR #474, opened 2024-08-30
Remove nullifying of service comment list before removal of a comment

jmacario24:
This pull request is to remove the nullifying of service comment list prior the removal of the comment from the list.

### DIFF
--- a/src/naemon/comments.c
+++ b/src/naemon/comments.c
@@ -154,7 +154,6 @@ int delete_comment(int type, unsigned long comment_id)
 	}
 	else if (type == SERVICE_COMMENT) {
 		service *temp_service = find_service(this_comment->host_name, this_comment->service_description);
-		temp_service->comments_list = NULL;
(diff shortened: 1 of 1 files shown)


nook24:
I guess this is fine as we do not nullify `temp_host->comments_list` either.

Could you please add one or two sentence about this issue this causes? Like did Naemon crash or something?

jmacario24:
> I guess this is fine as we do not nullify `temp_host->comments_list` either.
> 
> Could you please add one or two sentence about this issue this causes? Like did Naemon crash or something?

Trying to access the comment list after a first delete will return an empty list as can be verified by call to number_of_service_comments after a delete.

sni:
thanks for the clarification, i'll have a look. No idea why i did that 2 years ago...

sni (approved):

sni:
thanks for submitting this PR. I guess it was a test leftover.

PR #477, opened 2024-09-12
Feature global notification handler

sni:
Add global notification handler in the same way as global event handlers. It adds 2 new
global options:

- global_host_notification_handler=global-host-notify
- global_service_notification_handler=global-service-notify

which will be run each time a normal notification would be considered. It does not require
contacts for hosts and services. If used it will also prevent warnings if hosts/services do
not have contacts set.
The global notification handlers do not have any contact macros set, because they are independent of contacts.

This PR also adds another global option:

- log_global_notifications

This option can be used to suppress a log entry each time the global notification handler is run.

Finally this PR adds a new generic external command

- LOG

It can be used to add custom log entries to the main naemon log file. For example to add the
actual notified contacts to the log file.

ex.:

    printf "[%lu] LOG;SERVICE NOTIFICATION: test contact;testhost;test service;CRITICAL;notify-by-email;test message\n" `date +%s` >> .../tmp/run/naemon.cmd

which would add a log entry similar to a normal notification. The actual hosts/services/contacts and commands don't have to exist.

Or other notes like ex.:

    printf "[%lu] LOG;SERVICE NOTE: testhost;event handler restarted service successfully\n" `date +%s` >> .../tmp/run/naemon.cmd


src/naemon/configuration.c
```diff
@@ -1455,7 +1497,7 @@ int pre_flight_object_check(int *w, int *e)
 		}
 
 		/* check to see if there is at least one contact/group */
-		if (temp_service->contacts == NULL && temp_service->contact_groups == NULL) {
+		if (temp_service->contacts == NULL && temp_service->contact_groups == NULL && global_service_event_handler == NULL) {
```
nook24, line 1500: That should be probably `global_service_notification_handler`

src/naemon/defaults.h
```diff
@@ -47,6 +47,7 @@
 #define DEFAULT_LOG_CURRENT_STATES				1	/* log current service and host states after rotating log */
 #define DEFAULT_LOG_EXTERNAL_COMMANDS				1	/* log external commands */
 #define DEFAULT_LOG_PASSIVE_CHECKS				1	/* log passive service checks */
+#define DEFAULT_log_global_notifications			1	/* log global notifications */
```
nook24, line 50: Can we uppercase this please

src/naemon/logging.h
```diff
@@ -39,6 +39,7 @@
 
 #define NSLOG_HOST_NOTIFICATION		524288
 #define NSLOG_SERVICE_NOTIFICATION	1048576
+#define NSLOG_EXT_CUSTOM			2097152
```
nook24, line 42: Can you please remove one `\t` so all numbers line up nicely

src/naemon/notifications.c
```diff
@@ -2245,3 +2252,188 @@ int add_notification(notification **notification_list, nagios_macros *mac, conta
...
+	nj = nm_calloc(1, sizeof(struct notification_job));
+	nj->ctc = NULL;
+	nj->hst = hst;
+	nj->svc = NULL;
+	if (ERROR == wproc_run_callback(processed_command, notification_timeout, notification_handle_job_result, nj, mac)) {
+		nm_log(NSLOG_RUNTIME_ERROR, "wproc: Unable to send notification for host '%s' to worker\n", hst->name);
+		free(nj);
+	}
+
+	/* @todo Handle nebmod stuff when getting results from workers */
```
nook24, line 2426: Is this still relevant?

src/naemon/configuration.c
sni: thanks, you are totally right

src/naemon/defaults.h
sni: i am pretty sure it was uppercase initially :-) Will fix it again

src/naemon/logging.h
sni: looks aligned too me with ts=4. But i can replace all tabs in that file with spaces, but that'll be an extra commit.

src/naemon/notifications.c
sni: removed it...

src/naemon/logging.h
nook24: Looks like every editor has it's own opinion on that.
It's sort of aligned on the GitHub Website, but in VS Code the numbers are all over the place.

I guess your suggestion is the best option :)

grafik

PR #479, opened 2024-09-27
add hint of previous object when duplicate detection fails

sni:
whenever naemon encounters a 'Error: Duplicate definition found for' it will now print the location of the previous object definition as well to make debugging easier.

While on it, rework xodtemplate_add_object_property() a bit and clean up code. Instead of setting a result, simply return early with the error. There is no cleanup at the end or something like that. Also cleanup the double use of the result variable which was used for the result but also for determing if there is a value set or not.


sni:
adding and adopting #478 here since they had conflicts which are resolved here

nook24 (approved):
I also have checked for any leftovers but you already catched all boolean values :)

PR #481, opened 2024-09-30
enable codeql scanner & fix detected defects

sni:


sni:
I assume the codeql check fails because it is not yet enabled. This will happen with this PR.

nook24 (approved):

PR #484, opened 2024-11-11
cleanup output

sni:
only print the requested information.
when running with -V simply print the version and exit.
when running with -h only print help and exit
when running with -L only print license and exit

### DIFF
--- a/src/naemon/naemon.c
+++ b/src/naemon/naemon.c
@@ -125,6 +125,7 @@ int main(int argc, char **argv)
 {
 	int result;
 	int error = FALSE;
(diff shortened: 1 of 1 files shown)


nook24 (approved):

nook24:
Just for documentation purpose: with this change, the output of `naemon --version` will **not** start with a leading line break anymore.  (indeed no output does this anymore now)

Old output
```

Naemon Core 1.4.2
```

New output
```
Naemon Core 1.4.2
```

Just in case this breaks someone's parser.

PR #490, opened 2025-01-28
postpone check until check period starts

sni:
Given a service with a 24h check interval and a office hours check period. If this service is scheduled for whatever reasons outside office hours, it will be rescheduled 24hours again which is also outside office hours.
In the end, this service will never be checked again.

So, when naemon detects a check which could not be run because it is outside the check period, postpone the check until the next slot in its check period.


nook24 (approved):
Good catch!

PR #491, opened 2025-02-07
add support for configurable relative configuration paths

sni:
Naemon already supports relative include paths and uses the folder of the main naemon.cfg as base path.

This is fine unless you want to generate the main naemon.cfg and start naemon from a tmp folder. In this case, i'd like to set the base path from the config file.

By default nothing changes, but you can now override this by setting `config_rel_path`.

### DIFF
--- a/sample-config/naemon.cfg.in
+++ b/sample-config/naemon.cfg.in
@@ -11,7 +11,7 @@
 
 
 # BASE DIRECTORY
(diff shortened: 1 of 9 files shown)


nook24 (commented):
Can you also please add an example directive for `config_rel_path` into the naemon.cfg

nook24 (approved):
I have added this to the new docs :)
grafik

sni:
perfect, thanks

PR #494, opened 2025-02-20
improve postpone checks outside of check period

sni:
with #490 we rescheduled checks found outside their check period to the next time slot in their check period with a random delay of 60 seconds (or what ever retained_scheduling_randomize_window was set to)

There are 2 scenarios when this leads to scheduling issues:

1) Lots of checks with a medium check interval (ex. 2h) and office hours
   timeperiod. Those checks should evenly be scheduled over the 2h interval
   but currently, they would start with the office hours and from then they
   all at once every 2 hours which results in load peaks.
   The solution here is to take the check period into account when postponing
   the next check.
2) Consider checks with a long check interval (ex.: 24h) and small timeperiods,
   ex.: only 08:00 till 08:05. In this case we need to take the actual time slot
   into account to find a valid next check time slot.

While on it, i merged the code into a generic function which is then used for hosts and services.

### DIFF
--- a/src/naemon/checks.c
+++ b/src/naemon/checks.c
@@ -15,6 +15,7 @@
 #include "flapping.h"
 #include "sehandlers.h"
 #include "notifications.h"
+#include "objects_timeperiod.h"
 #include "logging.h"
 #include "globals.h"
 #include "nm_alloc.h"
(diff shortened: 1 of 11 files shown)


nook24 (approved):
I think the way you implemented it should also work for distributed setups, as long as the users make sure to keep `additional_freshness_latency` > `retained_scheduling_randomize_window` in the naemon.cfg

PR #514, opened 2026-01-19
fix memory leak when using service parents

sni:
add missing free to service parents.


sni:
OBS build issues fixed with #515

PR #526, opened 2026-03-16
Remove tv_str and use a unsigned int for slot in mkstr

nook24:
We have noticed random segfaults on one of our systems during the creating of `status.dat`.
Since Naemon 1.5.0, `tv_str` is called in the For-loop for each host and service status object, which will increase the counter of `slot` by one each time. After 2.147.483.647 the integer will flip into -2.147.483.648 and modulo of a negativ dividend will result in a negativ value in C (and PHP, Go JavaScript).
When reading `buf[-2147468803]` whatever is at this address will be overwritten.

### DIFF
--- a/lib/nsutils.c
+++ b/lib/nsutils.c
@@ -68,11 +68,6 @@ float tv_delta_f(const struct timeval *start, const struct timeval *stop)
 	return ret;
 }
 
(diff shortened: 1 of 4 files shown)


nook24:
Please see https://github.com/naemon/naemon-core/pull/527

Issue #530, opened 2026-03-28
Cannot build 1.5.1


merdely:
Hi

I'm trying to build 1.5.1 on Arch Linux and I'm getting this in my output:

```
make[1]: Entering directory '/tmp/naemon-naemon-core-611a5b6'
  CC       src/naemon/naemon-naemon.o
  CC       src/naemon/broker.lo
  CC       src/naemon/checks.lo
src/naemon/checks.c: In function 'parse_output':
src/naemon/checks.c:114:19: error: assignment discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
  114 |                 p = strpbrk((const char *) tmp, "|");
      |                   ^
src/naemon/checks.c:141:19: error: assignment discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
  141 |                 p = strpbrk((const char *) tmp, "|");
      |                   ^
cc1: all warnings being treated as errors
make[1]: *** [Makefile:2009: src/naemon/checks.lo] Error 1
make[1]: Leaving directory '/tmp/naemon-naemon-core-611a5b6'
make: *** [Makefile:1355: all] Error 2
```

I'm not sure of the RIGHT way to fix it.  I was able to get the warnings to go away if I do this:

```
p = strpbrk(tmp, "|");
```

But there are more instances of this structure throughout the code and I'm not even sure if this is right.
```
bang = (char *)strchr(name, '!');
```

Help, please

nook24:
It looks like this issue is related to https://github.com/naemon/naemon-core/issues/522, and I was able to reproduce this on my side as well.

```
[root@a8a35a7d2a79 naemon-core-1.5.1]# lsb_release -a
LSB Version:	n/a
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a

[root@a8a35a7d2a79 naemon-core-1.5.1]# gcc --version
gcc (GCC) 15.2.1 20260209
Copyright (C) 2025 Free Software Foundation, Inc.
```

It's interesting that it is failing now, even if the GCC version claims to be the same as I was successfully able to build it in this comment: https://github.com/naemon/naemon-core/issues/522#issuecomment-3912681401 a few weeks ago.

What ever. From my understanding of the error
```
src/naemon/checks.c:114:19: error: assignment discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
  114 |                 p = strpbrk((const char *) tmp, "|");
      |                   ^
[...]

Issue #522, opened 2026-02-16
Build failure with glibc 2.43


micwoj92:
Hello,
Trying to build naemon 1.5.0 (1.4.4 fails too) with gcc 15.2.1 20260209 fails with following error message:
```
src/naemon/checks.c: In function 'parse_output':
src/naemon/checks.c:114:19: error: assignment discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
  114 |                 p = strpbrk((const char *) tmp, "|");
      |                   ^
src/naemon/checks.c:141:19: error: assignment discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
  141 |                 p = strpbrk((const char *) tmp, "|");
      |                   ^
cc1: all warnings being treated as errors
make[1]: *** [Makefile:2009: src/naemon/checks.lo] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/build/naemon/src/naemon-git'
make: *** [Makefile:1355: all] Error 2
```

nook24:
Could it be possible that you are missing a required dependency or so?
I successfully build Naemon in a Docker container (`docker run --rm -it archlinux:latest bash`)

Arch version
```
[root@b7f20b030fe4 naemon-core-1.5.0]# cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
VERSION_ID=20260208.0.488728
```

gcc version
```
gcc (GCC) 15.2.1 20260209
```

Steps to reproduce

```
pacman -Sy  base-devel wget curl vim glib2-devel help2man gperf gcc  gdb  pkgconfig libtool

wget https://github.com/naemon/naemon-core/archive/refs/tags/v1.5.0.tar.gz
tar xfv v1.5.0.tar.gz
cd naemon-core-1.5.0/

./autogen.sh
./configure --prefix=/usr \
            --bindir=/usr/bin \
            --datadir="/usr/share/naemon" \
            --libdir="/usr/lib/naemon" \
            --localstatedir="/var/lib/naemon" \
            --sysconfdir="/etc/naemon" \
            --with-pkgconfdir="/etc/naemon" \
            --mandir="/usr/share/man" \
            --with-pluginsdir="/usr/lib/monitoring-plugins" \
            --with-tempdir="/var/cache/naemon" \
            --with-checkresultdir="/var/cache/naemon/checkresults" \
[...]

PR #531, opened 2026-03-29
Fix: Discarded-qualifiers errors in string searching functions

nook24:
Currently Naemon uses a mix of (char \*) and (const char \*) as input for different string functions such as strpbrk() or strchr(). Modern compilers and stricter build flags (-Werror) throw errors when the result of string searching functions like strpbrk() or strchr() is assigned to a non-const pointer, especially when the input was cast to const char\*.

This commit adds explicit (char \*) casts to the return values of strpbrk and strchr to ensure compatibility across different compiler versions and to satisfy -Werror=discarded-qualifiers.

Affected functions:
- strpbrk
- strchr

Issues:
- #530
- #522

### DIFF
--- a/lib/nspath.c
+++ b/lib/nspath.c
@@ -29,7 +29,7 @@ static inline int path_components(const char *path)
 	int comps = 1;
 	if (!path)
 		return 0;
-	for (slash = strchr(path, '/'); slash; slash = strchr(slash + 1, '/'))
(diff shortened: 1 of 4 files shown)


merdely (approved):
This fixed the build for me!  Thank you!

sni (approved):

PR #538, opened 2026-08-17
Reject boolean object directives with excess characters

lausser:
xod_parse_bool() examined only the first character of a directive's value, so everything after it was silently discarded. "register 0v" and "register 0nono" were read as 0, and "register 10" as 1.

The 0-prefixed case is the damaging one. The object is treated as an unregistered template, so a single stray character removes a host from monitoring while naemon starts up cleanly and reports nothing at all.

The value is now compared in full and only "0" and "1" are accepted, as the documentation has always specified. Anything else is a config error, so the object config is rejected, the pre-flight check does not run and naemon does not start.

The error also names the directive as written in the config file and quotes the offending value:

    Error: invalid value '0v' for 'register', expected 0 or 1.

It previously printed the struct member name, which for "register" is "register_object" - a directive that appears in no config file. The name now comes from the parsed token rather than the member, so aliases are reported as written too: "checks_enabled" is not reported as "active_checks_enabled", nor "obsess_over_host" as "obsess".

This is a breaking change. Configurations accepted before will now be rejected, so run naemon -v before upgrading and before reloading a running instance.

Affected directives:
- register
- active_checks_enabled (alias checks_enabled)
- passive_checks_enabled
- event_handler_enabled
- flap_detection_enabled
- check_freshness
- process_perf_data
- obsess (aliases obsess_over_host, obsess_over_service)
- is_volatile
- notifications_enabled
- host_notifications_enabled, service_notifications_enabled
- retain_status_information, retain_nonstatus_information
- inherits_parent
- can_submit_commands

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,20 @@
+Next:
+Breaking changes:
+* boolean object directives now accept only the exact values 0 and 1. Values carrying excess
+  characters, such as `register 0v` or `register 10`, were previously accepted silently: only the
+  first character was examined, so `0v` was read as 0 and the object was quietly left unregistered
+  and unmonitored. Such values are now a configuration error and Naemon will refuse to start.
+  This affects `register`, `notifications_enabled`, `active_checks_enabled` (alias `checks_enabled`),
+  `passive_checks_enabled`, `event_handler_enabled`, `flap_detection_enabled`, `check_freshness`,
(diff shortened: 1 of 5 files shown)


sni:
Changed this into a warning. The advantage is users will then see all issues at once.

nook24 (approved):

PR #542, opened 2026-09-12
Add dev container for building and debugging Naemon in Docker

nook24:
With the help of Claude AI, I added support to build and debug Naemon including Broker Modules inside dev containers.

Currently, it is only possible to develop Naemon on native Linux (or using WSL2 on Windows, which is a VM and therefore native linux). Most of the time, I'm not having a Linux system available and also I wanted to be more portable with my dev environment. To goal was to have a "One Click Solution" like we currently have on Linux with the VS Code Config.

Dev Containers are using Docker to achieve this. All you need on your host system is Docker and VS Code. Please see the readme for more details.

In addition, this setup can also be used to build and debug Naemon Event Broker Modules. Currently mod_gearman and the Statusengine broker are supported (any fully optional). You can set breakpoints in the Naemon Core and in the Broker Modules, just as you would do on Linux.


sni:
tbh, i am not a huge fan of putting each dev environment in the source repository.
ex.: there is https://github.com/sni/naemon-dev-box which is a separate project and it does mostly the same thing. It prepares a dev environment with naemon, livestatus, mod-gearman, etc...
Usually i'd try to avoid having lots of "unrelated" dotfiles in the source tree. Every couple of months there is a new `.<editor>` or `.<ci system>` and `.<super useful project>` which is forgotten after some time.

nook24:
Fair point. This could be a _me_ problem. Whenever I'm going to review a PR, I clone it, build it, run Naemon and go through the changes. That was the biggest motivation for me to automate the dev setup, for all platforms not just Linux.

The `.devcontainer` directory is an open standard [supported by](https://containers.dev/supporting) multiple tools and IDEs. But I see your point not to clutter the repo all over the place.

> ... which is forgotten after some time.

I could address this with an CI job, which builds the container and runs `make check`. This way it can not rot silently. And in case I'm not around anymore to fix it myself, you could drop it from the repository without breaking anything else.

**On naemon-dev-box:**
The `naemon-dev-box` is a full OMD stack which contains Grafana, Prometheus and more. My goal was to get a lightweight container running Naemon with an attached debugger and the capability loading/debugging broker modules.

**On moving it out of the repo:** that is possible, and VS Code has [documented](https://code.visualstudio.com/docs/devcontainers/create-dev-container#_alternative-repository-configuration-folders) this:

> In some cases, you may want to create a configuration for a repository that you do not control **or that you would prefer didn't have a configuration included in the repository itself**.

It uses the path of `git remote -v` to locate the configuration. Anyone working from a fork
needs to adjust the path manually. Which defeats the purpose.

I would rather keep it here with a CI job, but I could it also move it into a separate repo if you prefer.

nook24:
I'm closing this in favor of: https://github.com/naemon/naemon-core/pull/543

